PR #5, opened 2018-02-06
set content-type for subject

sharonovd:


smtp/init.lua
```diff
@@ -154,8 +154,13 @@ curl_mt = {
             if opts.headers and #opts.headers > 0 then
                 header = header .. table.concat(opts.headers, '\r\n') .. '\r\n'
             end
+
+            local content_type = opts.content_type or 'text/plain'
+            local charset = opts.charset or 'UTF-8'
             if opts.subject then
-                header = header .. 'Subject: ' .. opts.subject .. '\r\n'
+                header = header ..
+                    'Content-Type: ' .. content_type .. '; charset=' .. charset .. '\r\n' ..
```
Totktonada: Hmm, you set content type only if subject provided?

sharonovd: You are right, this does not make sense anymore. Moved.

PR #12, opened 2019-02-19
fix broken build on curl > 7.62.0

zwirec:
Use CURLE_SSL_CACERT in case value only if libcurl have version < 7.62.0 because in newer version CURLE_PEER_FAILED_VERIFICATION is alias for CURLE_SSL_CACERT as a result we have a broken build

### DIFF
--- a/smtp/smtpc.c
+++ b/smtp/smtpc.c
@@ -243,7 +243,9 @@ smtpc_execute(struct smtpc_request *req, double timeout)
 		req->status = (int) longval;
 		req->reason = "Ok";
 		break;
-	case CURLE_SSL_CACERT:
+#if LIBCURL_VERSION_NUM < 0x073e00
+	case CURLE_SSL_CACERT: /* deprecated in libcurl 7.62.0 */
+#endif
 	case CURLE_PEER_FAILED_VERIFICATION:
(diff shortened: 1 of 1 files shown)


Totktonada:
Please, see https://github.com/tarantool/tarantool/commit/74e20755f5a6ee506124b96a967c4ebf72e588ee and https://github.com/tarantool/tarantool/commit/02da15f7109b8bb7921eb97a751607b9dd2885b2

zwirec:
Okay, I updated my changes as recommended

smtp/smtpc.c
```diff
@@ -243,7 +243,9 @@ smtpc_execute(struct smtpc_request *req, double timeout)
 		req->status = (int) longval;
 		req->reason = "Ok";
 		break;
+#if LIBCURL_VERSION_NUM < 0x073e00
 	case CURLE_SSL_CACERT:
```
Totktonada: Nit: I would add `/* deprecated in libcurl 7.62.0 */` here as in https://github.com/tarantool/tarantool/commit/02da15f7109b8bb7921eb97a751607b9dd2885b2

Totktonada (approved):
Ok except minor nitpicking.

PR #14, opened 2019-09-02
Add curl to target_link_libraries

olegrok:
The problem was appeared then I try to install this module for tarantool with statically bundled libcurl


Totktonada (approved):
I have no objections. Please, verify that it works with one of tarantool's docker images and merge.

olegrok:
Works fine for tarantool with statically linked curl and without

Issue #11, opened 2019-01-24
Attachment handling


SWSAmor:
We need to send attachments. A new parameter table would be comfy. Do you have any plans for that?
Or maybe the attachment could be put in the body concatenated. Do you have any examples for that?
Thx.

Totktonada:
I didn't try, but with appropriate header and body parts in base64 it should be possible, consider https://stackoverflow.com/questions/44728855/curl-send-html-email-with-embedded-image-and-attachment

SWSAmor:
Thx working that way.

Totktonada:
Anyway it would be good to be able to handle this case via API.

PR #19, opened 2020-10-24
added api interface to attach files to email

no1seman:
subj.

full description of api achanges in README.md and in init.lua comments
Closing: https://github.com/tarantool/smtp/issues/11

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -1,0 +1,42 @@
+include_files = {"**/*.lua", "*.rockspec", "*.luacheckrc"}
+exclude_files = {"lua_modules", ".luarocks", ".rocks", ".history","test"}
+
+max_line_length = 120
+
+std = {
+    read_globals = {
+        '_G',
+        'arg',
+        'assert',
+        'box',
+        'collectgarbage',
+        'debug',
+        'dofile',
+        'error',
+        'getmetatable',
+        'io',
+        'ipairs',
+        'loadstring',
+        'math',
+        'next',
+        'os',
+        'package',
+        'pairs',
+        'pcall',
+        'print',
+        'rawget',
+        'rawset',
+        'require',
+        'select',
+        'setfenv',
+        'setmetatable',
+        'string',
+        'table',
+        'tonumber',
+        'tonumber64',
+        'tostring',
+        'type',
+        'unpack',
+        'xpcall',
+    },
+}
--- a/README.md
+++ b/README.md
@@ -46,13 +46,13 @@ You have two ways to install `tarantool/smtp`:
 
    With Tarantool 1.7.4+, say:
 
-   ```
+   ```bash
    tarantoolctl rocks install smtp
    ```
 
    With earlier Tarantool versions, set up Lua rocks and then say:
 
-   ```
+   ```bash
    luarocks --local install smtp
    ```
 
@@ -127,21 +127,57 @@ Example: `"Test Message"`.
 * `verbose` (boolean) -- whether `libcurl` verbose mode is enabled
 * `username` (string) -- a username for server authorization
 * `password` (string) -- a password for server authorization
+* `attachments` (table) -- a table (array) with attachments data
+  * `body` (any) attachment body contents
+  * `content_type` (string) -- set a content type (part of a Content-Type header,
+  defaults to 'text/plain')
+  * `charset` (string) -- set a charset (part of a Content-Type header, defaults
+  to 'UTF-8')
+  * `filename` (string) -- a string with filename will be shown in e-mail
+  * `base64_encode` (boolean) -- a boolean to base64 encode attachment content or not
 
 Example: `{timeout = 2}`
 
 Example of a complete request:
 
-```
+```lua
 response =
-client:request("smtp://127.0.0.1:34324",`"sender@tarantool.org"`,`"receiver@tarantool.org"`,"Test
-Message",{timeout=2})
+client:request("smtp://127.0.0.1:34324","sender@tarantool.org","receiver@tarantool.org","Test Message",{timeout=2})
 ```
 
 The response to the request will be a table containing a status (number)
 and a reason (string).
 Example: `{status: 250, reason: Ok}`
 (The standard status code 250 means the request was executed.)
+
+Example of a complete request with attachments:
+
+```lua
+response =
+client:request(
+  "smtp://127.0.0.1:34324",
+  "sender@tarantool.org",
+  "receiver@tarantool.org",
+  "Test Message",
+  {
+    timeout=2,
+    attachments = {
+      {
+          body = json.encode('{"key1":"value1"}'),
+          content_type = 'application/json',
+          charset = 'UTF-8',
+          filename = 'json1.json',
+          base64_encode = true
+      },
+      {
+          body = json.encode('{"key2":"value2"}'),
+          content_type = 'application/json',
+          filename = 'json2.json',
+          base64_encode = false
+      }
+    }
+  })
+```
 
 [Back to contents](#contents)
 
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -127,22 +127,58 @@ Example: `"Test Message"`.
...
 * `username` (string) -- a username for server authorization
 * `password` (string) -- a password for server authorization
+* `attachments` (table) -- a table (array) with attachments data
+  * `body` (any) attachment body contents
+  * `content_type` (string) -- set a content type (part of a Content-Type header,
+  defaults to 'text/plain')
+  * `charset` (string) -- set a charset (part of a Content-Type header, defaults
+  to 'UTF-8')
+  * `filename` (string) -- a string with filename will be shown in e-mail
+  * `base64_encode` (boolean) -- a boolean to base64 encode attachment content or not
```
Totktonada, line 137: What is default value?

README.md
```diff
@@ -127,22 +127,58 @@ Example: `"Test Message"`.
...
+          content_type = 'application/json',
+          charset = 'UTF-8',
+          filename = 'json1.json',
+          base64_encode = true
+      },
+      {
+          body = json.encode('{"key2":"value2"}'),
+          content_type = 'application/json',
+          filename = 'json2.json',
+          base64_encode = false
```
Totktonada, line 176: Non-encoded JSON in the strange example. I guess the `base64_encode = false` variant is useful, when you already have a base64 encoded data.

README.md
```diff
@@ -293,13 +329,13 @@ is done with a local test server, with none of the usual authorization options.)
...
 
-```
+```lua
 mails:get()
 ```
 
 And now the response should look like this:
 
-```
+```tarantool
```
Totktonada, line 338: Nit: There is no 'tarantool' language support on GitHub. I would use 'lua' or 'yaml'.

smtp/init.lua
```diff
@@ -55,13 +55,13 @@ local smtp_new = function(opts)
...
-local function check_args(self, method)
-    if type(self) ~= 'table' then
-        error(check_args_fmt:format(method, method), 2)
-    end
-end
+-- local function check_args(self, method)
+--     if type(self) ~= 'table' then
+--         error(check_args_fmt:format(method, method), 2)
+--     end
+-- end
```
Totktonada, line 64: It is better to remove the unused code: it's in the git history if anyone will need it (the same for `smtp_default` below). And It is better to move the dead code clean up into its own commit: this way it is easier to look over the history of changes.

smtp/init.lua
```diff
@@ -39,7 +39,7 @@ local curl_mt
 --
 --  Parameters:
 --
---  max_connectionss -  Maximum number of entries in the connection cache */
+--  max_connections -  Maximum number of entries in the connection cache */
```
Totktonada, line 42: I would move all those comment tweaks to its own commit (except ones that describe the functionality you added). It would be easier to look read the main changes this way.

smtp/init.lua
```diff
@@ -167,15 +181,60 @@ curl_mt = {
                 header = header .. table.concat(opts.headers, '\r\n') .. '\r\n'
             end
 
-            local content_type = opts.content_type or 'text/plain'
-            local charset = opts.charset or 'UTF-8'
-            header = header .. 'Content-Type: ' .. content_type .. '; charset=' .. charset .. '\r\n'
+            local attachments = ''
```
Totktonada, line 184: Nit: It is used only inside the `else` branch below.

smtp/init.lua
```diff
@@ -167,15 +181,60 @@ curl_mt = {
...
+                                  content_disposition ..
+                                  attachment_body
+                end
+
+                local content_type = 'Content-Type: ' .. (opts.content_type or 'text/plain') ..
+                                         '; charset=' .. (opts.charset or  'UTF-8') ..';\r\n\r\n'
+
+                if opts.subject then
+                    header = header .. 'Subject: ' .. opts.subject .. '\r\n' .. ''
+                end
```
Totktonada, line 227: I would perform an attempt to reduce code duplication where possible: either by changing of the code flow or by moving a common functionaliy into functions. Say, a function like `encode_content_type()` may be used as for the message itself as well as for attachments.

test/test.sh
```diff
@@ -0,0 +1,8 @@
+#/bin/bash
```
Totktonada, line 1: 1. Nit: There is not bash-specific here (any bourne shell compatible interpreter may run this code), so I would use `#!/bin/sh`. And, yep, the exclamation mark was forgotten.
2. I usually name such files with `.example` postfix and add a name without `.example` to gitignore. So a user may keep the repository clean and have secrets in a gitignored file. See token.txt.example [here](https://github.com/Totktonada/closed_issues) for example.
3. I would not name it 'test', because it does not verify correctness of the implementation. (The same for attach.test.lua.) I'll propose a name if something good will come into my mind.

smtp/init.lua
```diff
@@ -167,15 +181,60 @@ curl_mt = {
...
+                local charset = opts.charset or 'UTF-8'
+                header = header .. 'Content-Type: ' .. content_type .. '; charset=' .. charset .. '\r\n'
+
+                if opts.subject then
+                    header = header .. 'Subject: ' .. opts.subject .. '\r\n'
+                end
+
+                body = header .. '\r\n' .. body
+            else
+                local MULTIPART_CONTENT_TYPE = 'Content-Type: multipart/mixed; boundary=MULTIPART-MIXED-BOUNDARY;\r\n'
```
Totktonada, line 196: It would be good to mention exact standard you follow (at least because they're renewed from time to time).

Totktonada:
> adeded

Typo (in the commit message). I would also use imperative mood for the commit message header ('add') and mention the issue in the commit message body. See more recommendations [here](https://www.tarantool.io/en/doc/latest/dev_guide/developer_guidelines/#how-to-write-a-commit-message).

README.md
no1seman: OK, let's base64_encode will be true by default. I think base64_encode == false will be useful when need to send many tiny plain/text attachments. I think that option will be available.

.luacheckrc
```diff
@@ -0,0 +1,42 @@
...
+        'string',
+        'table',
+        'tonumber',
+        'tonumber64',
+        'tostring',
+        'type',
+        'unpack',
+        'xpcall',
+    },
+}
```
olegrok, line 42: There is no empty line at the end of file. Also seems you could use reduced version of .luacheckrc (see https://github.com/tarantool/cartridge/blob/master/.luacheckrc)

README.md
Totktonada: Resolved in PR #23.

README.md
Totktonada: Okay.

README.md
Totktonada: Resolved in PR #22.

smtp/init.lua
Totktonada: Partially resolved in PR #23, but it's up to Yaroslav.

smtp/init.lua
Totktonada: Resolved in PR #21.

smtp/init.lua
Totktonada: Resolved in PR #23.

smtp/init.lua
Totktonada: Mostly resolved in PR #23.

test/test.sh
Totktonada: The file was replaced by the actual test in PR #23, so the comment is not actual anymore.

smtp/init.lua
Totktonada: Resolved in PR #23.

.luacheckrc
Totktonada: The file was excluded from PR #23, so the coment becomes not actual.

PR #20, opened 2020-10-27
Check request input params

no1seman:
We have to check input params as early as we can to avoid of execution of any logic and also calling low-level functions if we exactly know that such params will cause an error. 
This little patch checks: "to" and "opts.cc" and "opts.bcc" is not nil, not an empty string or not an empty list. This is not garanties that at least one valid recipient present in params but checks most of possible errors.

### DIFF
--- a/smtp/init.lua
+++ b/smtp/init.lua
@@ -151,10 +151,10 @@ curl_mt = {
         --
         request = function(self, url, from, to, body, opts)
             opts = opts or {}
-            to = to or {}
-            if not body or not url or not from then
+            if not body or not url or not from or not (to or opts.cc or opts.bcc) then
                 error('request(url, from, to, body [, options]])')
             end
+            to = to or {}
             local header = ''
             local recipients = {}
             header = 'From: ' .. from .. '\r\n' ..
--- a/smtp/init.lua
+++ b/smtp/init.lua
@@ -151,10 +151,15 @@ curl_mt = {
         --
         request = function(self, url, from, to, body, opts)
             opts = opts or {}
-            if not body or not url or not from or not (to or opts.cc or opts.bcc) then
+            to = to or {}
+            if not body or not url or not from then
                 error('request(url, from, to, body [, options]])')
             end
-            to = to or {}
+            if (not to or not (to and to == {} or #to > 0)) and
+               (not opts.cc or not (opts.cc and opts.cc == {} or #opts.cc > 0)) and
+               (not opts.bcc or not (opts.bcc and opts.bcc == {} or #opts.cc > 0)) then
+                error('No recipients in request')
+            end
             local header = ''
             local recipients = {}
             header = 'From: ' .. from .. '\r\n' ..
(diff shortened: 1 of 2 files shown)


smtp/init.lua
```diff
@@ -151,10 +151,10 @@ curl_mt = {
         --
         request = function(self, url, from, to, body, opts)
             opts = opts or {}
-            to = to or {}
-            if not body or not url or not from then
+            if not body or not url or not from or not (to or opts.cc or opts.bcc) then
```
Totktonada: So `nil` and `{}` for `to` will be interpreted differently. That's strange.

Totktonada: The error that is given in the case does not help to spot the usage problem. I would give its own error message in the case.

Totktonada:
> If no any of: "to" or "cc" or "bcc" is set then infinite loop in request function is possible. This patch fixes that case.

Can you include some description into a commit message too?

smtp/init.lua
no1seman: Fixed, added tests for both nil, empty string and empty table cases

no1seman: Fixed

smtp/init.lua
```diff
@@ -155,6 +155,11 @@ curl_mt = {
             if not body or not url or not from then
                 error('request(url, from, to, body [, options]])')
             end
+            if (not to or not (to and to == {} or #to > 0)) and
+               (not opts.cc or not (opts.cc and opts.cc == {} or #opts.cc > 0)) and
+               (not opts.bcc or not (opts.bcc and opts.bcc == {} or #opts.cc > 0)) then
```
Totktonada, line 160:
```
tarantool> {} == {}
---
- false
...
```

The standard trick for checking a table for emptiness is `next(my_table) == nil`. However `#to > 0` should be enough in the case.

test/smtp.test.lua
```diff
@@ -82,6 +82,29 @@ test:test("smtp.client", function(test)
     m = mails:get()
     test:is_deeply(m.rcpt, {'<cc@tarantool.org>'}, 'no rcpt')
 
+    local ok = pcall(function() return
+    client:request(addr, 'sender@tarantool.org',
+                       nil,
+                       'mail.body',
+                       {})
+                    end)
```
Totktonada, line 90: The formatting is a bit crazy.

1. `return` on a separate line.
2. The nested code block is on the same indentation level as the parent one.
3. Arguments are indented in some unusual way (I guess due to the unusual indentation of `end`).

I would do this so:

```lua
local ok, err = pcall(function()
    return client:request(addr, 'sender@tarantool.org', nil, 'mail.body')
end)
test:is_deeply({ok, err}, {false, exp_err}, 'No recipients')
```

Or, better:

```lua
local ok, err = pcall(client.request, client, addr, 'sender@tarantool.org',
                      nil, 'mail.body')
```

Totktonada:
Well, let's return here, when you'll be ready to a dialogue.

PR #23, opened 2020-10-27
Add attachment handling

### LINKED ISSUE #11: Attachment handling

no1seman:
Add attachments handling api to lib. Full description in README.md and also some usefull comments in init.lua

Closes https://github.com/tarantool/smtp/issues/11

### DIFF
--- a/smtp/init.lua
+++ b/smtp/init.lua
@@ -31,6 +31,7 @@
 
 local fiber = require('fiber')
 local driver = require('smtp.lib')
+local digest = require('digest')
 
 local curl_mt
 
@@ -53,14 +54,6 @@ local smtp_new = function(opts)
 
     local curl = driver.new(opts.max_connections)
     return setmetatable({ curl = curl, }, curl_mt )
-end
-
-local check_args_fmt = 'Use client:%s(...) instead of client.%s(...):'
-
-local function check_args(self, method)
-    if type(self) ~= 'table' then
-        error(check_args_fmt:format(method, method), 2)
-    end
 end
 
 --
@@ -113,6 +106,21 @@ end
 --
 --      password - a password for server authorization;
 --
+--      attachments - a table with array of attachments
+--
+--          body - attachment body
+--
+--          content_type - set a content type (part of a Content-Type header,
+--              defaults to 'text/plain'), MIME type according to RFC2045
+--
+--          charset - set a charset (part of a Content-Type header, defaults to
+--          'UTF-8')
+--
+--          filename - a string with filename will be shown in e-mail
+--
+--          base64_encode - a boolean to base64 encode attachment content or not, defaults to true
+--
+--
 --  Returns:
 --      {
 --          status=NUMBER,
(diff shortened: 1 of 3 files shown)


smtp/init.lua
```diff
@@ -29,8 +29,8 @@
 --  SUCH DAMAGE.
 --
 
-local fiber = require('fiber')
```
Totktonada, line 32: I still think that all those deletions contradict with the approach, when we trying to keep commits atomic. It is easy to extract them into its own commit using `git add -p`. However, okay, up to you.

smtp/init.lua
```diff
@@ -155,27 +162,65 @@ curl_mt = {
...
             end
+            if opts.subject then
+                header = header .. 'Subject: ' .. opts.subject .. '\r\n'
+            end
             add_recipients(recipients, opts.bcc)
             if opts.headers and #opts.headers > 0 then
                 header = header .. table.concat(opts.headers, '\r\n') .. '\r\n'
             end
+            local content_type = 'Content-Type: ' .. (opts.content_type or 'text/plain') ..
+            '; charset=' .. (opts.charset or  'UTF-8') ..';\r\n'
```
Totktonada, line 181: Nit: This header is added only when there are no attachments, so I would sink it down to the corresponding if-branch.

smtp/init.lua
```diff
@@ -155,27 +162,65 @@ curl_mt = {
...
+                local MULTIPART_CONTENT_TYPE = 'Content-Type: multipart/mixed; boundary=MULTIPART-MIXED-BOUNDARY;\r\n'
+                local MULTIPART_SEPARATOR = '\r\n--MULTIPART-MIXED-BOUNDARY\r\n'
+                local MULTIPART_END = '\r\n--MULTIPART-MIXED-BOUNDARY--\r\n'
+                local attachments = ''
 
-            local content_type = opts.content_type or 'text/plain'
-            local charset = opts.charset or 'UTF-8'
-            header = header .. 'Content-Type: ' .. content_type .. '; charset=' .. charset .. '\r\n'
+                for _, attachment in ipairs(opts.attachments) do
+                    local base64_encode = attachment.base64_encode or true
```
Totktonada: `false or true` is `true`.

smtp/init.lua
```diff
@@ -155,27 +162,65 @@ curl_mt = {
...
+                    local attachment_content_type = 'Content-Type: ' ..
+                                         (attachment.content_type or 'text/plain') ..
+                                         '; charset=' ..
+                                         (attachment.charset or  'UTF-8') ..
+                                         ';\r\n'
+                    local content_transfer_encoding = attachment.base64_encode
```
Totktonada, line 199: We can use `base64_encode` variable here.

Totktonada (commented):
I'm okay in general, but there are several comments around. Most important: let's look at the case when `base64_encode` is set to `false`. It is invalid now: encoding is still performed, only `Content-Transfer-Encoding` header is omitted.

smtp/init.lua
no1seman: I'm disagree with your point cause 'content_type' is alo used in "with attachments" clause of if, see: lines 216-223 of init.lua
Generally in case of multipart bode content we still need point body content-type as it is a part of of multipart and every single part have to have content_type.

smtp/init.lua
no1seman: Fixed. Also I improved test suite to test both three cases: base64_encode = nil, false, true

smtp/init.lua
Totktonada: I overlooked it. Agreed.

Totktonada:
Squashed commits, added `Closes #11`.

Totktonada (approved):
LGTM.

PR #25, opened 2021-01-25
build: link against Tarantool's libCURL

### LINKED ISSUE #24: Support linking against tarantool's built-in libcurl
* [x] Enable SMTP support in tarantool's bundled libcurl.
  - Export all necessary libcurl symbols from tarantool executable (done in https://github.com/tarantool/tarantool/issues/4559).
  - Ship curl's public headers within `/usr/include/tarantool/curl` (done in https://github.com/tarantool/tarantool/issues/4559).
* [x] Support building of the module against just tarantool, without system libcurl, under a cmake option (update: implemented unconditionally).
* [x] Automatically determine whether given tarantool installation have all necessary symbols and headers to use it instead of system libcurl (update: use tarantool headers if available, but find necessary symbols in runtime).

romanhabibov:
Link against Tarantool's libCURL if it is possible (in the case of
libCURL included as bundled library or static build).

Closes #24

To test it, use Tarantool branches: [2.7](https://github.com/tarantool/tarantool/tree/romanhabibov/curl-smtp) and [1.10](https://github.com/tarantool/tarantool/tree/romanhabibov/curl-smtp_1_10).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,9 @@ set(TARANTOOL_FIND_REQUIRED ON)
 find_package(Tarantool)
 include_directories(${TARANTOOL_INCLUDE_DIRS})
 
+option(TARANTOOL_CURL_REQUIRED "Use Tarantool bundled lubCURL." OFF)
+option(USE_TARANTOOL_CURL "Use Tarantool bundled lubCURL." ON)
+
 # Set CFLAGS
 # set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
 set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -Wall -Wextra")
--- a/smtp/CMakeLists.txt
+++ b/smtp/CMakeLists.txt
@@ -4,7 +4,7 @@ endif(APPLE)
 
 # Add C library
 add_library(lib SHARED lib.c smtpc.c)
-target_link_libraries(lib curl ${MSGPUCK_LIBRARIES})
+target_link_libraries(lib curl)
 set_target_properties(lib PROPERTIES PREFIX "" OUTPUT_NAME "lib")
 
 # Install module
--- a/smtp/CMakeLists.txt
+++ b/smtp/CMakeLists.txt
@@ -4,7 +4,26 @@ endif(APPLE)
 
 # Add C library
 add_library(lib SHARED lib.c smtpc.c)
-target_link_libraries(lib curl)
+
+find_path(TARANTOOL_CURL_INCLUDE_DIR
+          NAMES curl/curl.h
+          PATH_SUFFIXES tarantool)
+if (TARANTOOL_CURL_INCLUDE_DIR AND USE_TARANTOOL_CURL)
+    message(STATUS "Link against tarantool's bundled lubCURL.")
+    set(CURL_LIBRARIES)
+    set(CURL_INCLUDE_DIRS "${TARANTOOL_CURL_INCLUDE_DIR}")
+elseif (TARANTOOL_CURL_REQUIRED)
+    message(FATAL_ERROR "Your Tarantool version doesn't have bundled lubCURL. "
+            "The minimum required version is 1.10.9 or 2.8.1.")
+else()
+    message(WARNING "Installed tarantool version does not ship lubCURL headers."
+            " Fallback to linking with system lubCURL.")
+    set(CURL_LIBRARIES curl)
+    set(CURL_INCLUDE_DIRS)
+endif()
+target_include_directories(lib PRIVATE ${CURL_INCLUDE_DIRS})
+target_link_libraries(lib PRIVATE ${CURL_LIBRARIES})
+
 set_target_properties(lib PROPERTIES PREFIX "" OUTPUT_NAME "lib")
 
 # Install module
--- a/debian/control
+++ b/debian/control
@@ -5,7 +5,6 @@ Maintainer: Roman Tsisyk <roman@tarantool.org>
 Build-Depends: debhelper (>= 9), cdbs,
                cmake (>= 2.8),
                tarantool-dev (>= 1.6.8.0),
-               libmsgpuck-dev (>= 1.0.0),
 # For /usr/bin/prove
                perl (>= 5.10.0),
                libcurl4-openssl-dev | libcurl4-gnutls-dev | libcurl4-nss-dev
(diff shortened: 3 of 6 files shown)


CMakeLists.txt
```diff
@@ -12,6 +12,8 @@ set(TARANTOOL_FIND_REQUIRED ON)
 find_package(Tarantool)
 include_directories(${TARANTOOL_INCLUDE_DIRS})
 
+option(USE_TARANTOOL_CURL "Use Tarantool bundled lubCURL." ON)
```
olegrok, line 16: typo: `lub` -> `lib`

smtp/CMakeLists.txt
```diff
@@ -4,7 +4,24 @@ endif(APPLE)
...
+            libCURL.")
+    target_link_libraries(lib curl ${MSGPUCK_LIBRARIES})
+  else()
+    message("-- Use Tarantool bundled libCURL.")
+    target_include_directories(lib PRIVATE
+                               ${HAS_BUNDLED_LIBCURL}/${TARANTOOL_CURL_PATH})
+  endif()
+else()
+  target_link_libraries(lib curl ${MSGPUCK_LIBRARIES})
+endif()
```
Totktonada: Let's start from usage scenarious:

1. A developer of the module can either install libcurl or use bundled one. Any behaviour is okay.
2. An application developer can install the module using luarocks / tarantoolctl rocks. The module will be used with currently installed tarantool. It is not redistributable to other systems. However updating of tarantool to a newer version should not break the module.
3. Building of a package. It seems, it will be okay to always link against the bundled libcurl. I guess such build will work as for tarantool with bundled libcurl as well as for one with the system one (except, maybe, cases with bundled libcurl without smtp(s) support).

So we need the following ways to build:

* Default: prefer libcurl bundled into tarantool, but fallback to a system libcurl.
* For packages: always use bundled libcurl; bump tarantool and tarantool-devel/dev dependecies.

So packages should pass an option like `-DTARANTOOL_CURL_REQUIRED=ON` to cmake (and the option should be disabled by default).

Next, I don't see your testing plan and results of the testing. Please, elaborate. Otherwise we'll duplicate effort.

I would not duplicate `target_link_libraries()` calls, but rather set `${CURL_LIBRARIES}` and `${CURL_INCLUDE_DIRS}`. Your way is error-prone. For example, one of branches missed `target_link_libraries()` at all, so the dependency on msgpuck is not set.

Hm... It seems, msgpuck is not used in the module, so, please, remove it first (in a separate commit) from `debian/control` and `smtp/CMakeLists.txt`. That's confusing.

I would write it like so (I didn't test):

```cmake
find_path(TARANTOOL_CURL_INCLUDE_DIR
    NAMES curl/curl.h
    PATH_SUFFIXES tarantool)
if (TARANTOOL_CURL_INCLUDE_DIR)
    message(STATUS "Link against tarantool's bundled libcurl.")
    set(CURL_LIBRARIES)
    set(CURL_INCLUDE_DIRS "${TARANTOOL_CURL_INCLUDE_DIR}")
elseif (TARANTOOL_CURL_REQUIRED)
    # XXX: Mention a tarantool versions necessary to succussfully
    # build the module.
    message(FATAL_ERROR "<...>")
else()
    message(WARNING "Installed tarantool version does not ship libcurl headers. Fallback to linking with system libcurl.")
    set(CURL_LIBRARIES curl)
    set(CURL_INCLUDE_DIRS)
endif()
target_include_directories(lib PRIVATE ${CURL_INCLUDE_DIRS})
target_link_libraries(lib PRIVATE ${CURL_LIBRARIES})
```

romanhabibov: I tested current branch with [2.8](https://github.com/tarantool/tarantool/tree/romanhabibov/curl-headers) and [1.10](https://github.com/tarantool/tarantool/tree/romanhabibov/curl-smtp_1_10):
developer's build with bundled curl works fine.
I didn't test static build and rocks installation.

Totktonada:
I propose the following testing plan:

1. Linux, the module from a package.
    - old tarantool
    - new tarantool
2. Linux, the module from `tarantoolctl rocks`.
    - old tarantool
    - new tarantool
3. Mac OS (tarantool from brew, without bundled libcurl), the module from `tarantoolctl rocks`.
    - old tarantool
    - new tarantool

We should build a module package against a new tarantool (after your patches). I guess we can achieve this by building necessary tarantool package and preinstall it in `{rpm,debian}/prebuild.sh` (but we should verify that it works as expected).

Totktonada:
Pushed the first commit (msgpuck removal).

smtp/CMakeLists.txt
Totktonada: It seems, we unable to support all 2.5-2.7 releases without a **runtime** logic, whether to use tarantool's built-in libcurl or system's one. I'll share details below.

Regarding this particular commit:

1. I would remove the `TARANTOOL_CURL_REQUIRED=ON` option from the rockspec. Packages are prebuilt, but installation using luarocks / tarantoolctl rocks assumes the build step. So the additional requirement for presence of the bundled libcurl in tarantool would forbid installation of the module on all existing ('old') tarantool versions.
2. I don't see a usage scenario for `-DUSE_TARANTOOL_CURL=OFF -DTARANTOOL_CURL_REQUIRED=OFF`, that's why I proposed the sole `-DTARANTOOL_CURL_REQUIRED=ON/OFF`. It seems, it was not obvious from my wording, sorry. (And the options are not orthogonal this way: `-DUSE_TARANTOOL_CURL=OFF -DTARANTOOL_CURL_REQUIRED=ON` looks useless and is not handled gracefully now). I would drop the `USE_TARANTOOL_CURL` option.
3. lubCURL/libCURL -> libcurl (see the explanation [here](https://lists.tarantool.org/pipermail/tarantool-patches/2021-March/023177.html)).
4. Without `NO_DEFAULT_PATH` the `find_path()` call may return a system's libcurl header, so we'll never fall into the `else()` branch.

Anyway, a build-time check does not look sufficient, because many tarantool releases in the 2.5-2.7 series expose libcurl functions without smtp(s) support. A function from the executable have precedence over one from system's libcurl even if we list /usr/lib/libcurl.so in the NEEDED field of the .so/dylib file.

Totktonada:
I written a short prove why a build time check would not be sufficient: https://github.com/tarantool/smtp/issues/24#issuecomment-822885014. In fact, we already have a problem with currently released tarantool's.

Totktonada:
Superseded by #32 and #33.

PR #26, opened 2021-03-11
Fix: subject RFC 822 compat

no1seman:
According to RFC 822#3.3 subject must contain characters with codes 0-127: https://tools.ietf.org/html/rfc822
For compatibility with legacy mail relays simply base64 encode subject field if char codes >127 is present

### DIFF
--- a/smtp/init.lua
+++ b/smtp/init.lua
@@ -152,6 +152,27 @@ local function addr_spec(mailbox)
     return mailbox:match('^.*[ \t]*<(.-)>[ \t]*$') or mailbox
 end
 
+local function is_gt_127(text)
+    for idx = 1, #text do
+        if text:byte(idx) > 127 then
+           return true
+        end
+    end
+    return false
+end
+
+-- According to RFC 822#3.3 subject must contain characters with codes 0-127:
+-- https://tools.ietf.org/html/rfc822
+-- For compatibility with legacy mail relays simply base64 encode subject field
+-- if char codes >127 is present
+local function encode_subject(subj)
+    if subj and type(subj) == 'string' and subj ~= '' and is_gt_127(subj) then
+        local encoded = digest.base64_encode(subj)
+        return '=?utf-8?b?' .. encoded .. '?='
+    end
+    return subj
+end
+
 curl_mt = {
     __index = {
         --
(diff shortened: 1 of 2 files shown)


sharonovd (approved):

smtp/init.lua
```diff
@@ -152,6 +152,29 @@ local function addr_spec(mailbox)
...
+            if text:byte(idx) > 127 then
+               return true
+            end
+        end
+        return false
+    end
+
+    if subj and type(subj) == 'string' and subj~='' and is_gt_127(subj) then
+        local encoded = digest.base64_encode(subj)
+        if encoded then
```
Totktonada: base64_encode() raises an error in case of an error. As I see, it never returns `nil`.

test/smtp.test.lua
```diff
@@ -114,20 +114,38 @@ test:test("smtp.client", function(test)
...
+                       'receiver@tarantool.org',
+                       'mail.body',{
+                           attachments = {
+                           {
+                               body = 'Test message',
+                               content_type = 'text/plain',
+                               filename = 'text.txt',
+                               base64_encode = true,
+                           }
+                       }})
```
Totktonada, line 126: Irrelevant code change.

Totktonada: Moved to its own commit.

smtp/init.lua
Totktonada: Removed the condition.

Totktonada (approved):
LGTM.

@no1seman Are you agree with the changes?

smtp/init.lua
```diff
@@ -152,6 +152,27 @@ local function addr_spec(mailbox)
     return mailbox:match('^.*[ \t]*<(.-)>[ \t]*$') or mailbox
 end
 
+-- According to RFC 822#3.3 subject must contain characters with codes 0-127:
+-- https://tools.ietf.org/html/rfc822
+-- For compatibility with legacy mail relays simply base64 encode subject field
+-- if char codes >127 is present
+local function encode_subject(subj)
+    local is_gt_127 = function(text)
```
olegrok: Do you really need to recreate this function for each call?

no1seman: Moved is_gt_127() out of encode_subject() scope

Totktonada: I squashed this commit into the first one.

olegrok (approved):

no1seman:
@Totktonada OK, I agree that additional nil check of base64_encode is redundant. Thank u for suggestion!

PR #27, opened 2021-03-14
Add new line after body header if attach present

no1seman:
Add absent new line between "content_type" and "body" if attachment present.

### DIFF
--- a/smtp/init.lua
+++ b/smtp/init.lua
@@ -181,7 +181,7 @@ curl_mt = {
             '; charset=' .. (opts.charset or  'UTF-8') ..';\r\n'
 
             if not opts.attachments or #opts.attachments == 0 then
-                body = header .. content_type ..'\r\n' .. body
+                body = header .. content_type .. '\r\n' .. body
             else
                 -- multipart content according to https://tools.ietf.org/html/rfc1341
                 local MULTIPART_CONTENT_TYPE = 'Content-Type: multipart/mixed; boundary=MULTIPART-MIXED-BOUNDARY;\r\n'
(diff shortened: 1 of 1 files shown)


smtp/init.lua
```diff
@@ -178,10 +178,10 @@ curl_mt = {
                 header = header .. table.concat(opts.headers, '\r\n') .. '\r\n'
             end
             local content_type = 'Content-Type: ' .. (opts.content_type or 'text/plain') ..
-            '; charset=' .. (opts.charset or  'UTF-8') ..';\r\n'
+            '; charset=' .. (opts.charset or  'UTF-8') ..';\r\n\r\n'
 
             if not opts.attachments or #opts.attachments == 0 then
-                body = header .. content_type ..'\r\n' .. body
+                body = header .. content_type .. body
```
Totktonada: I would add extra `'\r\n'` between `content_type` and `body` in the if-branch that handles attachments (around lines 219-220) to improve code durability. Now it is easy to miss that content-type should always going right before body. The patch to illustrate the idea:

```diff
diff --git a/smtp/init.lua b/smtp/init.lua
index 47c813d..94ae12a 100644
--- a/smtp/init.lua
+++ b/smtp/init.lua
@@ -238,6 +238,7 @@ curl_mt = {
                        header ..
                        MULTIPART_SEPARATOR ..
                        content_type ..
+                       '\r\n' ..
                        body ..
                        attachments ..
                        MULTIPART_END
```

no1seman: No, I disagree, seems it's better to add '\r\n' once in content_type and new line wiil be added in both cases.

no1seman (commented):
Ok, Fixed

Totktonada (approved):
Looks correct.

Do you test it somehow?

PR #28, opened 2021-03-14
Fix a various bugs related to attachments feature

no1seman:
This PR fixes some bugs which leds attachments doesn't work properly:

1. New line needed between body header and body payload;
2. Subject base64_encode must be a single line so nowarap opt added;
3. Content-Transfer-Encoding: base64 must be the last in attachment header and must not finished by ";";
4. filename attribute value in Content-Disposition header must be double quoted.

After that fixes base64 coded attachments finally working as expected, tested on: mail.ru, gmail.com and MacOS Mail client.

### DIFF
--- a/smtp/init.lua
+++ b/smtp/init.lua
@@ -167,7 +167,7 @@ end
 -- if char codes >127 is present
 local function encode_subject(subj)
     if subj and type(subj) == 'string' and subj ~= '' and is_gt_127(subj) then
-        local encoded = digest.base64_encode(subj)
+        local encoded = digest.base64_encode(subj, { nowrap = true })
         return '=?utf-8?b?' .. encoded .. '?='
     end
     return subj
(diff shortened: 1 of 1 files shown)


Totktonada (approved):
I don't mind.

Can you include the description of fixes (like you do in the PR) into the commit message as well?

smtp/init.lua
```diff
@@ -218,26 +218,27 @@ curl_mt = {
...
                                   content_disposition ..
+                                  content_transfer_encoding ..
                                   attachment_body
                 end
 
                 body = MULTIPART_CONTENT_TYPE ..
                        header ..
                        MULTIPART_SEPARATOR ..
                        content_type ..
+                       '\r\n'..
```
sharonovd: style - please add whitespace

sharonovd:
> I don't mind.
> 
> Can you include the description of fixes (like you do in the PR) into the commit message as well?

Agreed

Issue #29, opened 2021-04-21
ER_UNKNOWN: SMTP error 1 (os errno 0) on 2.5/2.6/2.7


Totktonada:
The 'error 1' is [CURLE_UNSUPPORTED_PROTOCOL](https://github.com/curl/curl/blob/6bdc96fee93cc23d9b059021f6d2e24bf8a8dc76/include/curl/curl.h#L484): there is no smtp(s) support in tarantool 2.5.{1,2,3}, 2.6.{1,2} and 2.7.1 ([fixed](https://github.com/tarantool/tarantool/issues/4559) in 2.6.3, 2.7.2 and 2.8.1; see [details](https://github.com/tarantool/smtp/issues/24#issuecomment-822885014)). It is important, because a dynamic loader [prefers](https://github.com/tarantool/tarantool/discussions/5733) symbols from the executable even when we link over system libcurl.

The only option I see is to choose libcurl implementation in runtime using dlopen() / dlsym().

PR #30, opened 2021-04-21
ci: add initial CI

Totktonada:
In future I plan to enhance it with testing of real world scenarious:
build packages for different OSes and verify it against different
tarantool version on each OS, verify the module installed from rocks
too.

However I hope this testing will be enough to have some evidence that an
upcoming fix for known linking problems (#29) resolves the problem.

Part of #29
Part of #24

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -1,0 +1,242 @@
+name: testing
+
+on: [push, pull_request]
+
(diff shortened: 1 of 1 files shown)


LeonidVas (approved):
I looked throught the patch. I don't have any experience of working with GHA. So if it works then it LGTM.

Totktonada:
Filed follow up issues: #34, #35, #36, #37.

PR #32, opened 2021-04-23
libcurl: load system libcurl library in runtime

### LINKED ISSUE #29: ER_UNKNOWN: SMTP error 1 (os errno 0) on 2.5/2.6/2.7

Totktonada:
It allows to avoid usage of libcurl functions provided by tarantool executable. Those functions are not usable from the module perspective, because there are tarantool releases, which are bundled with libcurl without smtp(s) support. Another problem is that there are releases that do not enforce exporting of the full list of libcurl symbols, so a linker is free to eliminate them from the executable.

See the list of affected tarantool releases in #24 comments. In short, they are 2.5.{1,2,3}, 2.6{1,2} and 2.7.1.

See details in the commit message.

This pull request also adds a check for protocols supported by libcurl and gives a human readable error message at `require('smtp')` if smtp or smtps is not supported.

Automatic choosing between tarantool's built-in libcurl and system's libcurl is out of scope of this pull request and will be implemented in the scope of #24.

Part of #24.
Fixes #29.

### DIFF
--- a/smtp/CMakeLists.txt
+++ b/smtp/CMakeLists.txt
@@ -4,7 +4,94 @@ endif(APPLE)
 
 # Add C library
 add_library(lib SHARED lib.c smtpc.c)
-target_link_libraries(lib curl)
+
+# We MUST NOT add the curl library here.
+#
+# Note: The problem is described from the user perspective in [1].
+# The text below is the long explanation how it was solved.
+#
+# The reason why we must avoid dynamic linking with system's
+# libcurl is composition of several facts.
+#
+# The module uses system's libcurl.so, because libcurl symbols
+# that are exported from tarantool executable are not always
+# suitable for the module.
+#
+# 1. There are tarantool releases that do not export libcurl
+#    symbols at all.
+# 2. There are tarantool releases that export partial list of
+#    libcurl public symbols (due to default visibility and lack
+#    of an explicit export list).
+# 3. There are tarantool releases, which offer libcurl without
+#    smtp(s) support.
+#
+# See the affected tarantool releases list in [2].
+#
+# Anyway, the module calls dlopen() with RTLD_DEEPBIND (see
+# smtpc_init()) for loading system's libcurl.so and everything
+# should be fine? Sadly, no...
+#
+# System's libcurl.so is usually built without
+# -Wl,-Bsymbolic-functions (the exception is Ubuntu -- it
+# surprisingly differs here from Debian) and dynamic relocations
+# are generated for exported libcurl functions that libcurl calls
+# from its code. Those calls are dynamic and an application can
+# retarget them just by exporting the target function from its
+# executable. A user may retarget those calls using LD_PRELOAD.
+# This mechanism aims to provide a way to extend or change a
+# shared library functionality.
+#
+# However at least some of such calls (from libcurl's code to a
+# libcurl's exported function) are internal by its nature and,
+# AFAIU, must land to the same libcurl.so version / build. The
+# example I found is calling of curl_multi_add_handle() from the
+# curl_easy_perform() implementation. It fails for me when
+# system's libcurl.so (7.75.0 with smtp(s) support) processes an
+# smtp request and calls curl_multi_add_handle() from tarantool
+# 2.5.3 (libcurl 7.71.1 without smtp(s) support). The error is
+# CURLE_FAILED_INIT.
+#
+# Back to the main question. What is going on if we'll add curl
+# library to the NEEDED entry of the module's dynamic library
+# (using `ld <...> -lcurl` or so)?
+#
+# Tarantool is loaded and calls require('smtp'), the module's
+# lib.so library starts to load. The dynamic linker find the
+# NEEDED entry with system's libcurl.so and loads it. The
+# libcurl.so library has dynamic relocations and they are resolved
+# to the executable's symbols. The module loads libcurl.so using
+# dlopen() that just increases a reference counter of already
+# loaded library. The resolved dynamic relocations remain resolved
+# to the executable and the RTLD_DEEPBIND flag does not affect
+# them. Due to differences in versions / supported protocols (I
+# don't know a presice reason) curl_easy_perform() from the loaded
+# libcurl.so becomes broken and always returns CURLE_FAILED_INIT.
+#
+# What if we'll not add the curl library here?
+#
+# If we're lucky and system's libcurl.so is not loaded without
+# RTLD_DEEPBIND yet, our dlopen() with RTLD_DEEPBIND will work
+# as expected and all symbols from the library will be resolved
+# to the library itself.
+#
+# Sure, if another module has system's libcurl.so in the NEEDED
+# entry or loads it without RTLD_DEEPBIND and this another module
+# is loaded prior to this one we again in the trouble.
+#
+# There are several facts that more or less mitigate the problem:
+#
+# 1. I don't know any other tarantool's module that uses libcurl
+#    (except built-in http.client, which uses built-in libcurl,
+#    and so irrelevant here).
+# 2. Tarantool offers suitable libcurl implementation since
+#    1.10.10, 2.6.3, 2.7.2, 2.8.1 and in all 2.9+. The module
+#    will use the built-in libcurl when possible (it is not
+#    implemented yet, tracked by gh-24).
+#
+# [1]: https://github.com/tarantool/smtp/issues/29
+# [2]: https://github.com/tarantool/smtp/issues/24
+target_link_libraries(lib ${CMAKE_DL_LIBS})
+
 set_target_properties(lib PROPERTIES PREFIX "" OUTPUT_NAME "lib")
 
 # Install module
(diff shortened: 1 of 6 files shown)


Totktonada:
To be pushed after #30.

smtp/smtpc.c
```diff
@@ -140,6 +163,14 @@ smtpc_init(void)
 	load_func(libname, libcurl_handle, curl_easy_strerror);
 	load_func(libname, libcurl_handle, curl_slist_append);
 	load_func(libname, libcurl_handle, curl_slist_free_all);
+	load_func(libname, libcurl_handle, curl_version_info);
+
+	/* Verify that given libcurl supports smtp(s). */
+	curl_version_info_data *info = curl_version_info(7);
```
LeonidVas: "age should  be  set to  the  version  of this functionality by the time you write your program. This way, libcurl will always return a proper  struct  that  your program understands, while programs in the future might get a different struct." Cool idea.

smtp/CMakeLists.txt
```diff
@@ -4,7 +4,90 @@ endif(APPLE)
...
+# 1. I don't know any other tarantool's module that uses libcurl
+#    (except built-in http.client, which uses built-in libcurl,
+#    and so irrelevant here).
+# 2. Tarantool offers suitable libcurl implementation since
+#    1.10.10, 2.6.3, 2.7.2, 2.8.1 and in all 2.9+. The module
+#    will use the built-in libcurl when possible (it is not
+#    implemented yet, tracked by gh-24).
+#
+# [1]: https://github.com/tarantool/smtp/issues/29
+# [2]: https://github.com/tarantool/smtp/issues/24
```
LeonidVas, line 92: Wow! Good research!

smtp/smtpc.c
```diff
@@ -34,10 +34,134 @@
...
+	load_func(libname, libcurl_handle, curl_slist_append);
+	load_func(libname, libcurl_handle, curl_slist_free_all);
+
+	return 0;
+}
+
+#undef load_func
+
+void
+smtpc_free(void)
```
LeonidVas: Are you planning to use this function anywhere?

LeonidVas (approved):
Generally LGTM. I suggest you to use `curl:` as a subsystem for the title of a commit message.
See some comments below.

smtp/smtpc.c
Totktonada: :)

smtp/CMakeLists.txt
Totktonada: Thanks! I plan to update several details soon.

smtp/smtpc.c
Totktonada: Good catch.

Not yet. Maybe in a future. It seems, I should leave a note in the commit message.

We have no any trigger on a module unloading. However I guess it is possible to add a GC handler on the module table. It is also possible to use [on_shutdown trigger](https://github.com/tarantool/tarantool/issues/5723) on recent tarantool versions to free resources before tarantool exit.

Possible use cases are hot update on system's libcurl.so (without restarting of tarantool; looks are quite rare need) and testing with a leak sanitiser (however I don't know whether it tracks dlopen() / dlclose() calls).

Totktonada: Added a paragraph into the commit message.

smtp/CMakeLists.txt
Totktonada: I clarified it a bit with mention of the reference counting that dlopen() performs. This way it should be easier to understand how all those things work.

Totktonada:
Rearranged the code a bit to shrink #33 diff.

Totktonada:
> I suggest you to use `curl:` as a subsystem for the title of a commit message.

Good idea, but I would prefer `libcurl:`. 'curl' is the name of the tool, but 'libcurl' or 'curl library' is the name of the library. Changed commit messages and the PR title.

Totktonada:
Rebased after #30 and #31 push.

smtp/smtpc.c
Totktonada: However just calling it from a GC handler would not be sufficient: the GC handler may be called after the second initialization call and will set libcurl function pointers to NULL. Since we're not actually experience any problem due to lack of deinitialization, I'll remove it.

I verified that hot reload works:

```diff
diff --git a/test/smtp.test.lua b/test/smtp.test.lua
index f46bcda..eae255f 100755
--- a/test/smtp.test.lua
+++ b/test/smtp.test.lua
@@ -7,6 +7,12 @@ local fiber = require('fiber')
 local socket = require('socket')
 local os = require('os')
 
+package.loaded['smtp'] = nil
+package.loaded['smtp.client'] = nil
+package.loaded['smtp.client.driver'] = nil
+package.loaded['smtp.lib'] = nil
+client = require('smtp').new()
+
 test:plan(1)
 mails = fiber.channel(100)
```

For the record, the description from the previous commit message:

> An attentive reader may find that smtpc_free() is not called from
> anywhere. Sure, we have no any trigger on unloading a module (on 
> `package.loaded.foo = nil`; and the module's table may be saved
> somewhere anyway). However it worth to keep the function up to date with
> smtpc_init() if we'll need it in future to implement hot reload of
> system's libcurl (without tarantool restart) or clean up a leak
> sanitizer results (don't know, however, whether it tracks dlopen() and 
> dlclose()). In theory it is possible to call it at garbage collecting of
> the module table (add a field with cdata with a metatable with a __gc
> metamethod). Or do it on tarantool shutdown using an atexit() trigger,
> tarantool's box.ctl.on_shutdown() or recently added C level
> on_shutdown() (see [3]). Or we can expose a function that a user may 
> call to fully deinitialize the module. Anyway, I don't see a need now,
> so I left it as possible future enhancement.

I replaced it with the following one:

> There is no smtpc_free() function that would complement smtpc_init(),
> because one extra loaded dynamic library is not a problem. I don't want
> to add code that does not solve any real problem. We can add it later if
> it'll be necessary to don't confuse some leak sanitizer.

The patch with the function removal:

```diff
diff --git a/smtp/smtpc.c b/smtp/smtpc.c
index 30e2eb1..f671a4e 100644
--- a/smtp/smtpc.c
+++ b/smtp/smtpc.c
@@ -39,7 +39,7 @@
 
 #include <module.h>
 
-/* {{{ Subsystem (de)initialization */
+/* {{{ Subsystem initialization */
 
 /* https://gcc.gnu.org/onlinedocs/gcc/Alternate-Keywords.html */
 #ifndef __GNUC__
@@ -119,19 +119,6 @@ bind_libcurl_functions(const char *libname, void *libcurl_handle)
 
 #undef load_func
 
-static void
-unbind_libcurl_functions(void)
-{
-	curl_easy_cleanup_ptr = NULL;
-	curl_easy_getinfo_ptr = NULL;
-	curl_easy_init_ptr = NULL;
-	curl_easy_perform_ptr = NULL;
-	curl_easy_setopt_ptr = NULL;
-	curl_easy_strerror_ptr = NULL;
-	curl_slist_append_ptr = NULL;
-	curl_slist_free_all_ptr = NULL;
-}
-
 int
 smtpc_init(void)
 {
@@ -171,18 +158,7 @@ smtpc_init(void)
 	return 0;
 }
 
-void
-smtpc_free(void)
-{
-	unbind_libcurl_functions();
-
-	if (libcurl_handle != NULL) {
-		dlclose(libcurl_handle);
-		libcurl_handle = NULL;
-	}
-}
-
-/* Subsystem (de)initialization }}} */
+/* Subsystem initialization }}} */
 
 int
 smtpc_env_create(struct smtpc_env *env)
diff --git a/smtp/smtpc.h b/smtp/smtpc.h
index d9099e4..a7a0c72 100644
--- a/smtp/smtpc.h
+++ b/smtp/smtpc.h
@@ -35,7 +35,7 @@
 
 #include <curl/curl.h>
 
-/* {{{ Subsystem (de)initialization */
+/* {{{ Subsystem initialization */
 
 /**
  * Initialize the subsystem.
@@ -48,16 +48,7 @@
 int
 smtpc_init(void);
 
-/**
- * Deinitialize the subsystem.
- *
- * Idempotent: may be called before smtpc_init() or after failed
- * initialization.
- */
-void
-smtpc_free(void);
-
-/* Subsystem (de)initialization }}} */
+/* Subsystem initialization }}} */
 
 /** {{{ Environment */
```

PR #33, opened 2021-04-23
libcurl: prefer libcurl from tarantool

### LINKED ISSUE #24: Support linking against tarantool's built-in libcurl

Totktonada:
The change is quite straightforward: attempt to find necessary functions
in tarantool's executable (or libraries it depends on) and verify that
they're suitable for the module. If they're not, issue a warning and try 
to load system's libcurl.

I left the dependency on system's libcurl in the RPM / Deb packages,
because there are tarantool releases on which it is required for loading
of the module.

Maybe it is possible to depend either on a new tarantool release or on
an old tarantool release plus system's libcurl, but I'll leave it as
possible future enhancement.

Building of the module from the rockspec does not require libcurl.so to
be present in the system. It also don't require presence of libcurl
headers in ${PREFIX}/include/curl if they're shipped by tarantool.

Presence of system's libcurl.so is not required to load the module on
recent tarantool releases: 1.10.10, 2.6.3, 2.7.2+, 2.8.1+, 2.9+ (when it
is built with bundled libcurl that is the case for our RPM / Deb 
packages). Tarantool installation on Mac OS from brew is linked with
system's libcurl and the module will use it too.

Aside of this, added a couple of checks into CI to verify that in 
presence of libcurl headers that are shipped by tarantool, system ones 
are not required, and that tarantool's ones are preferred when both are 
available.

Fixes #24

### DIFF
--- a/smtp/smtpc.c
+++ b/smtp/smtpc.c
@@ -153,12 +153,21 @@ bind_libcurl_functions(const char *libname, void *libcurl_handle)
 int
 smtpc_init(void)
 {
+	if (bind_libcurl_functions("[tarantool]", RTLD_DEFAULT) == 0)
+		return 0;
+
 	const char *libname;
 #ifdef __APPLE__
 	libname = "libcurl.dylib";
 #else
 	libname = "libcurl.so";
 #endif
+
+	/* Warn a user that we unable to use built-in libcurl. */
+	box_error_t *err = box_error_last();
+	const char *err_msg = box_error_message(err);
+	say_warn("%s", err_msg);
+	say_warn("Attempt to fallback to %s", libname);
 
 	int flags = RTLD_NOW | RTLD_LOCAL;
 	/*
@@ -186,6 +195,7 @@ smtpc_init(void)
 	if (bind_libcurl_functions(libname, libcurl_handle) != 0)
 		return -1;
 
+	say_warn("Successfully loaded %s", libname);
 	return 0;
 }
 
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -41,6 +41,20 @@ jobs:
           - live/2.7
           - live/2.8
           - live/2.9
+        # Add an extra check or modify a system environment
+        # on one job.
+        include:
+          # Applicable for: 2.6.3, 2.7.2+, 2.8.1+, 2.9+ (gh-4559).
+          #
+          # Verify that system's libcurl headers are not required.
+          - tarantool: live/2.8
+            dont_install_system_libcurl_header: true
+          # Applicable for: 2.6.3, 2.7.2+, 2.8.1+, 2.9+ (gh-4559).
+          #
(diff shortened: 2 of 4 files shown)


Totktonada:
In progress now. To be pushed after #32. Supersedes #25.

Totktonada:
Ready for review now. Based on #30 and #32, so includes its commits.

Totktonada:
Rebased after #30 and #31 push. To be rebased again after #32.

Totktonada:
Removed deinitialization in #32 (see the discussion [here](https://github.com/tarantool/smtp/pull/32#discussion_r619695122)), updated this PR with the same change.

smtp/smtpc.c
```diff
@@ -34,10 +34,132 @@
...
+#define curl_slist_append	curl_slist_append_ptr
+#define curl_slist_free_all	curl_slist_free_all_ptr
+
+/* dlopen() handle. Saved to call dlclose(). */
+static void *libcurl_handle;
+
+/*
+ * Set <...>_ptr using dlsym() call.
+ *
+ * Return -1 from the function that uses the macro on error (and
```
LeonidVas: Return**s**

smtp/smtpc.c
```diff
@@ -82,10 +84,31 @@ define_func_ptr(curl_slist_free_all)
...
+#define curl_version_info	curl_version_info_ptr
 
 /* dlopen() handle. Saved to call dlclose(). */
 static void *libcurl_handle;
 
+/**
+ * Check that given given protocol is supported according to
+ * given libcurl informational structure.
+ *
+ * Return 0 when the protocol is supported, -1 otherwise (and sets
```
LeonidVas: Return**s**

LeonidVas (approved):
LGTM.

smtp/smtpc.c
Totktonada: Those are two possible styles. I prefer the imperative mood for headings and short statements, when they don't form a real text. See [this dicsussion](https://stackoverflow.com/q/8910214/1598057) for example.

However, to be honest, I have no a strict formal rule about this.

smtp/smtpc.c
Totktonada: (Answered above.)

Totktonada:
Rebased after #32 push.

PR #38, opened 2021-04-27
ci: introduce new flow for rockspec pushing

### LINKED ISSUE #37: Deploy a rockspec from GitHub Actions
When scm-1 rockspec is changed or a new tag is pushed. See the [rocks.tarantool.org](https://github.com/tarantool/rocks.tarantool.org) repository.

Follows up #30.

olegrok:
This patch introduces Github Actions workflow to push rockspec.

Closes #37

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,36 @@
+name: publish
+
+on:
+  push:
+    branches: [master]
+    tags: ['*']
+
+jobs:
+  publish-scm-1:
+    if: github.ref == 'refs/heads/master'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: tarantool/rocks.tarantool.org/github-action@master
+        with:
+          auth: ${{ secrets.ROCKS_AUTH }}
+          files: smtp-scm-1.rockspec
+
+  publish-tag:
+    if: startsWith(github.ref, 'refs/tags/')
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      # Make a release
+      - run: printf '%s=%s\n' TAG "${GITHUB_REF##*/}" >> "${GITHUB_ENV}"
+      - run: sed -E
+          -e "s/branch = '.+'/tag = '${{ env.TAG }}'/g"
+          -e "s/version = '.+'/version = '${{ env.TAG }}-1'/g"
+          smtp-scm-1.rockspec >> smtp-${{ env.TAG }}-1.rockspec
+
+      - uses: tarantool/rocks.tarantool.org/github-action@master
+        with:
+          auth: ${{ secrets.ROCKS_AUTH }}
+          files: |
+            smtp-${{ env.TAG }}-1.rockspec


.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
+name: Publish
```
Totktonada: Nit: I would name it from a lowercase letter for consistency with the workflow we already have there.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
...
+
+on:
+  push:
+    branches: [master]
+    tags: ['*']
+
+jobs:
+  publish-scm-1:
+    if: github.ref == 'refs/heads/master'
+    runs-on: ubuntu-20.04
```
Totktonada: Since there is nothing specific for Ubuntu Focal, I would use just ubuntu-latest.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: tarantool/rocks.tarantool.org/github-action@master
+        with:
+          auth: ${{ secrets.ROCKS_AUTH }}
+          files: smtp-scm-1.rockspec
+
+  publish-tag:
+    if: startsWith(github.ref, 'refs/tags/')
+    runs-on: ubuntu-20.04
```
Totktonada: (Same here.)

.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
...
+    env:
+      CMAKE_LDOC_FIND_REQUIRED: 'YES'
```
Totktonada: ldoc is not used within this repository and there is no such variable anywhere.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: rosik/setup-tarantool@v1
```
Totktonada: (If we'll actually need it.)

1. rosik -> tarantool
2. `@v1` -> `@v2`

.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
...
+        with:
+          tarantool-version: '2.8'
+
+      # Setup ldoc
+      - run: tarantoolctl rocks install ldoc
+          --server=https://tarantool.github.io/LDoc/
+      - run: echo $PWD/.rocks/bin >> $GITHUB_PATH
```
Totktonada: (Same as above.)

.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
...
+
+      # Make a release
+      - run: echo TAG=${GITHUB_REF##*/} >> $GITHUB_ENV
```
Totktonada: 1. Prefer `printf` over `echo` in scripts (see [here](https://wiki.ubuntu.com/DashAsBinSh#echo)).
2. Always quote a variable expansion.
3. Prefer explicit curly brackets.

I would write it [this way](https://github.com/tarantool/smtp/blob/d748c55d9f1258896e1c431991726154162435a5/.github/workflows/testing.yml#L79-L82):

```sh
printf '%s=%s\n' TAG "${GITHUB_REF##*/}" >> "${GITHUB_ENV}"
```

Since it is used over several steps, I would move the variable assignment into its own step.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
...
+      - run: tarantoolctl rocks new_version --tag ${{ env.TAG }}
```
Totktonada: Just `sed` is definitely better. `new_version` strips comments, reformat everything, leave the `branch` field together with `tag` one. So it is more complex solution to produce worse result.

And I would keep quotes around any externally controlled value.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
...
+      - run: tarantoolctl rocks install smtp-${{ env.TAG }}-1.rockspec
```
Totktonada: Is it to check whether `s/branch = 'master'/tag = '<...>'` breaks anything? Hmm, don't know how it may occur.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
...
+      - run: tarantoolctl rocks pack smtp ${{ env.TAG }}
```
Totktonada: > please don't upload any platform-dependent `.*.rock`.

(From rocks.tarantool.org [README](https://github.com/tarantool/rocks.tarantool.org/blob/acf843e836020ee0bb6f4e00c94ac6c70158ffdb/README.md).)

AFAIU, it effectively means any module with built C code.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
...
+
+      - uses: tarantool/rocks.tarantool.org/github-action@master
+        with:
+          auth: ${{ secrets.ROCKS_AUTH }}
+          files: |
+            smtp-${{ env.TAG }}-1.rockspec
+            smtp-${{ env.TAG }}-1.all.rock
```
Totktonada: (Same as above.)

.github/workflows/publish.yml
Totktonada: If it only needed for `tarantoolctl rocks new_version/install/pack`, I would remove it and use `sed` (it is not the only reason to used sed, see below).

Totktonada (commented):
Thanks! Please, consider several comments below.

Totktonada:
(The secret was added.)

.github/workflows/publish.yml
olegrok: Fixed

.github/workflows/publish.yml
olegrok: Fixed

.github/workflows/publish.yml
olegrok: Fixed

.github/workflows/publish.yml
olegrok: Removed

.github/workflows/publish.yml
olegrok: tarantool/setup-tarantool removed at all

.github/workflows/publish.yml
olegrok: Removed

.github/workflows/publish.yml
olegrok: Removed

.github/workflows/publish.yml
olegrok: Removed

.github/workflows/publish.yml
olegrok: Yes, removed

.github/workflows/publish.yml
olegrok: It should be done to pack binary rock. Removed.

.github/workflows/publish.yml
olegrok: Fixed

.github/workflows/publish.yml
olegrok: Fixed

.github/workflows/publish.yml
```diff
@@ -0,0 +1,36 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      # Make a release
+      - run: printf '%s=%s\n' TAG "${GITHUB_REF##*/}" >> "${GITHUB_ENV}"
+      - run: sed -E
+          -e "s/branch = '.+'/tag = '${{ env.TAG }}'/g"
+          -e "s/version = '.+'/version = '${{ env.TAG }}-1'/g"
+          smtp-scm-1.rockspec >> smtp-${{ env.TAG }}-1.rockspec
```
Totktonada, line 30: Verified YAML syntax:

```
tarantool> print(require('yaml').decode(require('http.client').get('https://raw.githubusercontent.com/tarantool/smtp/a14240a73954897f7d48c2816345ba94bb9385ac/.github/workflows/publish.yml').body).jobs['publish-tag'].steps[3].run)

sed -E -e "s/branch = '.+'/tag = '${{ env.TAG }}'/g" -e "s/version = '.+'/version = '${{ env.TAG }}-1'/g" smtp-scm-1.rockspec >> smtp-${{ env.TAG }}-1.rockspec
```

Verified the sed expressions:

```diff
$ diff -u smtp-scm-1.rockspec <(sed -E -e "s/branch = '.+'/tag = '0.0.6'/g" -e "s/version = '.+'/version = '0.0.6-1'/g" smtp-scm-1.rockspec)
--- smtp-scm-1.rockspec	2021-04-19 21:24:57.075189726 +0300
+++ /dev/fd/63	2021-04-28 18:12:10.169930214 +0300
@@ -4,12 +4,12 @@
 -- version of the package; it's mandatory, but we don't use it in Tarantool;
 -- instead, provide below a specific branch in the package's repository at
 -- GitHub and set version to some stub value, e.g. 'scm-1'
-version = 'scm-1'
+version = '0.0.6-1'
 
 -- url and branch of the package's repository at GitHub
 source  = {
     url    = 'git://github.com/tarantool/smtp.git';
-    branch = 'master';
+    tag = '0.0.6';
 }
 
 -- general information about the package;
```

Everything works as expected, nice!

Totktonada (approved):
I don't see any problems visually. Let's try it so :)

Totktonada:
Thank you!

Totktonada:
[Success](https://github.com/tarantool/smtp/runs/2458951699)!

```sh
$ tarantoolctl rocks search smtp | grep 0.0.6
   0.0.6-1 (rockspec) - http://rocks.tarantool.org
```

Totktonada:
As stated in the comment https://github.com/tarantool/smtp/pull/38#discussion_r621464837 above, we should not upload the binary rock. However we can upload the source rock and speed up installation a bit. I implemented it in PR #41.

PR #42, opened 2021-08-03
ci: verify RPM/Deb packages

### LINKED ISSUE #34: Verify RPM / Deb packages using GitHub Actions
For each supported OS: build a package, install tarantool, install the module from the package and run tests.

Follows up #30.

romanhabibov:
Added GitHub workflow, which builds the module packages, installs these
packages and runs tests.

Enabled reduced list of distributions: only fresh/popular distributions.
It should be enough to catch most of real problems. Those workflows are 
not fast, so let's enable more targets if there will be a demand.

Resolved a package build problem on Fedora 33+. Resolved a package build 
problem on Ubuntu Bionic (it is not included into CI, however, because there is
newer Ubuntu Focal). See details in commit messages.

Fixes #34


Totktonada:
Rebased after PR #43.

Totktonada:
I polished it. In short:

* Made naming and style fixes, attempts to simplify the YAML file.
* Moved out package building fixes into its own commits.
* Replaced Linux distribution versions with more recents ones.
* Fixed build of a package for Fedora 33+.
* Worked around a problem with installer.sh on Debian/Ubuntu (it appears on Debian Bullseye).
* Replaced tarantool 1.10.10 with 1.10.7, found bug #44.
* Corrected package versions by using the deep clone (`fetch-depth: 0`).
* Enable skipping of pull request jobs, when the source branch is in the same repository.
* Explicitly set artifacts life time (`retention-days: 21`).
* Added tarantool version verification step (just in a case).
* Don't install module's source packages.
* Install a package with its dependencies on CentOS/Fedora (added TODO comment for Debian/Ubuntu).

Since I became a co-author now, my review is insufficient. I'll ask someone else to look on the changes.

LeonidVas (approved):
Hi! Thanks to the authors for the patchset.
Generally LGTM, but I have several questions:
* Why didn't you remove `.travis.yml`?
* Why are you using `live` repositories for tests if they are deprecated according to the new release policy (AFAIU)?

Totktonada:
> * Why didn't you remove `.travis.yml`?

Good point. Dropped `.travis.yml` and `Jenkinsfile`.

> * Why are you using `live` repositories for tests if they are deprecated according to the new release policy (AFAIU)?

Our intention was to use a latest tarantool version here and it was `live/2.9`. However now there is `pre-release/2` with tarantool-2.10.0-beta1. To be honest, I want to postpone this part (to don't do it in a hurry), filed #46.

The existing repositories are not deprected. The approach is deprecated: we'll not use it for newer repositories.

Totktonada (approved):

Issue #13, opened 2019-08-16
Any 5xx error from SMTP server becomes a strange internal error


akudiyar:
For example., the server (mock) is set up for returning a 510 code for some sort of e-mails. The server returns "510 Bad email address" in that case. 
On the client side, I want to see the exact response code and error message, but instead I get a runtime error with message "SMTP error 55 (os errno 0)".

The proposed changes to the client: check the response code and return it back with the error message
```
 long response;
 curl_easy_getinfo(handle, CURLINFO_RESPONSE_CODE, response); 
```

RunsFor:
I'm receiving runtime error from smtp client when server sends 3xx,4xx and 5xx status codes:
```
SMTP error 55 (os errno 61)
```

RunsFor:
If SMTP server responds with some weird message, i.e. `'1000 UNKNOWN ERROR\r\n'`, smtp client even **hangs up**.

PR #45, opened 2021-09-23
Add handing of SMTP server error messages.

dmitryalexeyev:
According to https://github.com/tarantool/smtp/issues/13 smtpc hangs if receives unexpected error code from smtp server or reports meaningless error messages if gets a.e. 5xx error codes.

This patch adds handling of curl errors even if smtp server drops connection. It also provides test for SMTP error code 510 handling.


test/smtp.test.lua
```diff
@@ -29,7 +29,11 @@ local function smtp_h(s)
             s:write('250 OK\r\n')
         elseif l:find('RCPT TO:') then
             mail.rcpt[#mail.rcpt + 1] = l:sub(9):sub(1, -3)
-            s:write('250 OK\r\n')
+	    if l:find('unexisting') then
```
Totktonada: Mix of tabs and spaces.

smtp/smtpc.c
```diff
@@ -440,6 +442,15 @@ smtpc_execute(struct smtpc_request *req, double timeout)
 		++env->stat.failed_requests;
 		smtpc_request_delete(req);
 		return -1;
+        case CURLE_SEND_ERROR:
+		{
+		char error_msg[CURL_ERROR_SIZE];
+		snprintf(error_msg, sizeof(error_msg), "SMTP error: %s", curl_error);
+		req->reason = (const char *)strndup(error_msg, CURL_ERROR_SIZE);
```
Totktonada: Here we leak the memory.

smtp/smtpc.c
```diff
@@ -440,6 +442,15 @@ smtpc_execute(struct smtpc_request *req, double timeout)
 		++env->stat.failed_requests;
 		smtpc_request_delete(req);
 		return -1;
+        case CURLE_SEND_ERROR:
```
Totktonada: Mix of tabs and spaces.

smtp/smtpc.c
```diff
@@ -440,6 +442,15 @@ smtpc_execute(struct smtpc_request *req, double timeout)
...
+		{
+		char error_msg[CURL_ERROR_SIZE];
```
Totktonada: > You must keep the associated buffer available until libcurl no longer needs it. Failing to do so will cause very odd behavior or even crashes. libcurl will need it until you call curl_easy_cleanup or you set the same option again to use a different pointer.

https://curl.se/libcurl/c/CURLOPT_ERRORBUFFER.html

We don't call curl_easy_cleanup() before returning from this stack frame.

smtp/smtpc.c
```diff
@@ -378,6 +378,7 @@ smtpc_task_execute(va_list list)
 int
 smtpc_execute(struct smtpc_request *req, double timeout)
 {
+	char curl_error[CURL_ERROR_SIZE];
```
Totktonada: So we'll allocate it on the stack always, even when there is no any need. It is 256 bytes, not much, so should be okay if we'll not do it always: it is just waste of stack space. Allocation of arrays on a fiber stack (which is 512 KiB, not default 8 MiB) is generally discouraged: however 256 bytes should be okay.

Totktonada: NB: We cannot convert it to the static array, because multiple fibers may enter into this function and yield on coio_call().

smtp/smtpc.c
Totktonada: We need the second buffer only to concatenate a constant string with given one, allocate a dynamic memory and copy the result to it? The second buffer looks redundant. Just allocate the dynamic memory and copy everything that is needed to it.

smtp/smtpc.c
Totktonada: I see now: we need to pass a buffer pointer to libcurl before start a request. So the reason above is not applicable. Sorry, I miss a concentration a bit tonight.

Anyway, we allocate it on the stack just to copy it into dynamic memory later. I think it would be better to allocate this within the request structure. This way we'll not need to do those copies. We'll not need to do the extra malloc() / strndup() below.

We also can pass `curl_error + sizeof("SMTP error: ")` to libcurl and fill the buffer start with those bytes when an error actually occurs. So we'll not need to do those concatenations using `snprintf()`.

test/smtp.test.lua
```diff
@@ -146,6 +150,11 @@ test:test("smtp.client", function(test)
                   m.text,
                   "Subject: =%?utf%-8%?b%?YWJjZGVmZ2hpamvRj2xtbm9wcXJzdHV2d3h5eg==%?=", ""))
     test:is(subj, 1, 'subject codes >127')
+
+    r = client:request(addr, 'sender@tarantool.org',
+                       'unexisting@tarantool.org',
+                       'mail.body')
+    test:is(r.reason, 'SMTP error: RCPT failed: 510', 'unexisting recipient')
```
Totktonada: How about 3xx and 4xx responses? How about malformed responses? Those cases are mentioned in the issue.

smtp/smtpc.c
```diff
@@ -378,6 +378,7 @@ smtpc_task_execute(va_list list)
 int
 smtpc_execute(struct smtpc_request *req, double timeout)
```
Totktonada: The commits formatting violates [our guidelines](https://www.tarantool.io/en/doc/latest/dev_guide/developer_guidelines/#how-to-write-a-commit-message):

> `1.` Separate subject from body with a blank line.
> `4.` Do not end the subject line with a period.

We also don't send fixup commits for review: nobody is interested in it. They should be squashed into appropriate commits, so a reader will always see a patchset in a clean state.

The commit message should close the issue like `Fixes #xxx` or `Closes #xxx`. See https://github.blog/2013-01-22-closing-issues-via-commit-messages/.

PR #47, opened 2021-09-28
Add handing of SMTP server error messages

### LINKED ISSUE #13: Any 5xx error from SMTP server becomes a strange internal error

dmitryalexeyev:
…mtp server or reports meaningless error messages if gets a.e. 5xx error codes.

This patch adds handling of curl errors if smtp server drops connection or doesn't reply correctly on rcpt command.
It also provides test for SMTP error codes 3xx, 4xx, 5xx and unexpected errors handling.

Fixes #13

### DIFF
--- a/smtp/smtpc.c
+++ b/smtp/smtpc.c
@@ -250,8 +250,17 @@ smtpc_request_new(struct smtpc_env *env, const char *url, const char *from)
 			      "Can't alloc curl handle");
 		return NULL;
 	}
+
+	req->curl_error = (char *) malloc(CURL_ERROR_SIZE);
+	if (req->curl_error == NULL) {
+		free(req);
+		box_error_set(__FILE__, __LINE__, ER_MEMORY_ISSUE,
+					  "Can't alloc curl buffer for error message");
+		return NULL;
+	}
 	curl_easy_setopt(req->easy, CURLOPT_URL, url);
 	curl_easy_setopt(req->easy, CURLOPT_MAIL_FROM, from);
+	curl_easy_setopt(req->easy, CURLOPT_ERRORBUFFER, req->curl_error);
 
 	return req;
 }
@@ -260,6 +269,9 @@ static long int
 smtpc_task_delete(va_list list)
 {
 	struct smtpc_request *req = va_arg(list, struct smtpc_request *);
+	if (req->curl_error != NULL) {
+		free(req->curl_error);
+	}
 	curl_easy_cleanup(req->easy);
 	return 0;
 }
@@ -438,8 +450,16 @@ smtpc_execute(struct smtpc_request *req, double timeout)
 		box_error_set(__FILE__, __LINE__, ER_MEMORY_ISSUE,
 			      "Curl internal memory issue");
 		++env->stat.failed_requests;
+		free(req->curl_error);
 		smtpc_request_delete(req);
 		return -1;
+
+	case CURLE_RECV_ERROR:
+	case CURLE_SEND_ERROR:
+		req->reason = req->curl_error;
+		req->status = -1;
+		++env->stat.failed_requests;
+		break;
 	default: {
 		char error_msg[256];
 		curl_easy_getinfo(req->easy, CURLINFO_OS_ERRNO, &longval);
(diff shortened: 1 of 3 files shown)


smtp/smtpc.c
```diff
@@ -438,14 +450,23 @@ smtpc_execute(struct smtpc_request *req, double timeout)
...
+		++env->stat.failed_requests;
+		break;
 	default: {
 		char error_msg[256];
 		curl_easy_getinfo(req->easy, CURLINFO_OS_ERRNO, &longval);
 		snprintf(error_msg, sizeof(error_msg), "SMTP error %i (os errno %li)", req->code, longval);
 		box_error_set(__FILE__, __LINE__, ER_UNKNOWN, error_msg);
 		++env->stat.failed_requests;
+		free(req->curl_error);
 		smtpc_request_delete(req);
```
Totktonada: It seems, we'll free `curl_error` twice in this case.

smtp/lib.c
```diff
@@ -78,7 +78,7 @@ luaT_smtpc_request(lua_State *L)
 	if (req == NULL)
 		return luaT_error(L);
 
-	double timeout = 365 * 24 * 3600;
+	double timeout = 10;
```
Totktonada: The default timeout is changed here for all usages, not just for tests.

smtp/smtpc.c
```diff
@@ -250,8 +250,17 @@ smtpc_request_new(struct smtpc_env *env, const char *url, const char *from)
 			      "Can't alloc curl handle");
 		return NULL;
 	}
+
+	req->curl_error = (char *) malloc(CURL_ERROR_SIZE);
```
Totktonada, line 254: Why don't allocate it as part of the request structure?

Totktonada:
(The comments above are just after slight glance, I didn't looked into the patch decently.)

It seems, it supersedes PR #45. Please, use force-pushes rather than creating a new PR for each new version of a patchset. This way we'll have the discussion in one place.

smtp/lib.c
dmitryalexeyev: Yes, I know this one isn't very good. But 10 seconds timeout for RCPT command seems reasonable to me, something wrong must be happening if we hit it.
Not sure about 1 year timeout om RCPT command, we need to bail out sooner or later if something went wrong. And we won't get to DATA cmd until we pass RCPT, so we hang forever here.
Here comes second issue - 10 seconds timeout works fine on mac, but seems to be ignored on Linux.

Shall we distinguish test case and normal work? Shall we use 1 year timeout on RCPT? Doubt it, probably 10 to 30 seconds would be better

dmitryalexeyev:
The problem is I deleted the repository, I originally created pr #45 from :(

smtp/lib.c
Totktonada: Sure, infinite timeouts are generally discouraged. But it is not the reason to change it here: it is irrelevant to the issue. And we definitely should not change it silently: without any mention in the commit message (but better move to its own commit).

AFAIU, we should pass the timeout option to test the fix of the error message and something went wrong with this functionality on Linux. If it is so, it looks as a bug, which blocks testing of this change. I suggest to file an issue about this with a small reproducer and tarantool/libcurl/module versions (plus platform: distribution and distro version).

dmitryalexeyev: I see; my bad. I missed the point I can control timeout from lua code directly.

> If it is so, it looks as a bug, which blocks testing of this change. I suggest to file an issue about this with a small reproducer and tarantool/libcurl/module versions (plus platform: distribution and distro version).

Sure, will do.

dmitryalexeyev: The issue is smtp_h spamming 502 Not implemented every time it receives empty reply, we just needed return statement after it. Not related to curl timeout in any way.
Fixed.

PR #48, opened 2021-10-18
ci: add reusable testing workflow

ylobankov:
The idea of this workflow is to be a part of the 'integration.yml'
workflow from the tarantool repo to verify the integration of the
smtp module with an arbitrary tarantool version.

This workflow is not triggered on a push to the repo and cannot be run
manually since it has only the 'workflow_call' trigger. This workflow
will be included in the tarantool development cycle and called by the
'integration.yml' workflow from the tarantool project on a push to the
master and release branches for verifying integration of tarantool with
smtp.

Part of tarantool/tarantool#6523
Part of tarantool/tarantool#5265
Part of tarantool/tarantool#6056

### DIFF
--- a/.github/workflows/reusable_testing.yml
+++ b/.github/workflows/reusable_testing.yml
@@ -1,0 +1,32 @@
+name: reusable_testing
+
+on:
+  workflow_call:
+    inputs:
+      artifact_name:
+        description: 'The name of the tarantool build artifact'
+        default: ubuntu-focal
+        required: false
+        type: string
+
+jobs:
+  run_tests:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: 'Clone the smtp module'
+        uses: actions/checkout@v2
+        with:
+          repository: ${{ github.repository_owner }}/smtp
+
+      - name: 'Download the tarantool build artifact'
+        uses: actions/download-artifact@v2
+        with:
+          name: ${{ inputs.artifact_name }}
+
+      - name: 'Install tarantool'
+        # TODO(ylobankov): Install package dependencies. Now we're lucky: all
+        # dependencies are already there.
+        run: sudo dpkg -i tarantool*.deb
+
+      - run: cmake . && make
+      - run: make check


ylobankov:
Related PR https://github.com/tarantool/tarantool/pull/6525

VitaliyaIoffe:
FYI, `smtp` now is using `openssl@3` on macos, but we haven't migrated even on the master branch. So, red ci could be fixed if you install previous openssl version

see [openssl issue](https://github.com/tarantool/tarantool/issues/6468)

Totktonada:
Thanks for the clarification of the Mac OS problem in CI, I filed #49 for that. Don't bother with it.

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: 'Install tarantool'
+        # TODO(ylobankov): Install package dependencies. Now we're lucky: all
+        # dependencies are already there.
+        run: sudo dpkg -i tarantool*.deb
+
+      - name: 'Install smtp'
+        run: tarantoolctl rocks make
+
+      - run: cmake .
+      - run: make check
```
Totktonada: Either build it using `tarantoolctl rocks make` (and run tests using `make -C build.luarocks check`) or build it using `cmake . && make` (and run tests using just `make check`). If you doing both actions, you just doing it twice.

Any of the ways is okay for me, but I would suggest to stick with the simpler one.

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,48 @@
...
+        with:
+          repository: ${{ github.repository_owner }}/smtp
+
+      - name: 'Download the tarantool build artifact'
+        uses: actions/download-artifact@v2
+        with:
+          name: ${{ inputs.artifact_name }}
+      # TODO(ylobankov): Just in case remove tarantool that could be previously
+      # installed on the runner. Relevant to self-hosted runners. Remove this
+      # step when only GitHub-hosted runners are used for integration testing.
```
Totktonada: It'll spread across ~20 repositories soon? And, if we'll need to adjust those workarounds, you'll open ~20 PRs? And wait for maintainers to look at them and push?

Can we spend a day of work and pin those runs to GitHub hosted runners? It'll be around 20 seconds or even less: almost nothing comparing to heavy tarantool's jobs.

VitaliyaIoffe: Have we decided that this is the best solution now? Are we definitely not changing it after a near period?
I'd like to understand this point and fix it for the future.

Totktonada: It looks as the best now, because reduces manual clean up activities as well as effort to automate them. What about a future: I hope, if we'll pay more effort to our own infra, we'll move toward autocreation of clean virtual machines. Otherwise, sooner or later, we'll completely miss understanding of the environment, where our workflows run.

Even if we'll return back here in a future and will want to run this workflow on self-hosted machines, I hope, they will not require such clean up.

ylobankov: Opened related PR https://github.com/tarantool/tarantool/pull/6537.

ylobankov: PR is merged. It will be deleted.

.github/workflows/reusable_testing.yml
ylobankov: Done.

VitaliyaIoffe (approved):

Totktonada (approved):
LGTM.

Issue #49, opened 2021-10-20
ci: 'Could NOT find OpenSSL' fails


Totktonada:
A month after 943a914664a3d18fc9bf897dd9bbba7b7f60b67f CI on Mac OS becomes broken again. Now it looks so:

```
-- Could NOT find OpenSSL, try to set the path to OpenSSL root folder in the system variable OPENSSL_ROOT_DIR (missing: OPENSSL_CRYPTO_LIBRARY OPENSSL_INCLUDE_DIR) 
CMake Error at CMakeLists.txt:315 (message):
  Could NOT find OpenSSL development files (libssl-dev/openssl-devel package)
```

The problem was worked around in tarantool in 1.10.11-8-g321c093cd, 2.8.2-18-gc4b649bba and 2.10.0-beta1-65-g338309788, see https://github.com/tarantool/tarantool/issues/6468. It seems, we should apply the same patch for affected tarantool versions in our CI.

PR #51, opened 2021-12-09
curl: unable to load libcurl.so

### LINKED ISSUE #44: Unable to load libcurl.so
## Affected versions

Affected versions of the module: 0.0.5+ (where we added the falling back to the system libcurl). Previous versions of the module may fail on those tarantool versions due to other reasons and with other disagnostics (see #24 and #29).

The problem appears on all tarantool versions, where the module falls back to system libcurl: all versions prior to 1.10.10, 2.6.3, 2.7.2, 2.8.1.

## How it looks

When run a script with `require('smtp')`:

```
Unable to load symbol curl_easy_cleanup from [tarantool]
Attempt to fallback to libcurl.so
ER_SYSTEM: Unable to load libcurl.so
fatal error, exiting the event loop
```

In the interactive console:

```yaml
tarantool> require('smtp')
Unable to load symbol curl_easy_cleanup from [tarantool]
Attempt to fallback to libcurl.so
---
- error: Unable to load libcurl.so
...
```

## When it appears and when not

The problem appears when the module falls back to the system libcurl (on relatively old tarantool versions) and when there is no dynamic library named exactly as 'libcurl.so' (without the version suffix). This situation usually occurs, when the module is installed from an RPM or Deb package and the developer libcurl package is not installed.

## Workarounds

While the problem is not fixed in the module, the following recipes should help to mitigate it.

### Install libcurl developer package (recommended)

The developer libcurl package provides libcurl.so -> libcurl.so.4.X.Y symlink.

On Debian or Ubuntu (verified on Debian Buster):

```sh
$ apt-get install libcurl4-openssl-dev
```

On CentOS or Fedora (verified on CentOS 7):

```sh
$ yum install libcurl-devel
```

### Create the symlink manually (alternative)

This option is not recommended.

On Debian or Ubuntu:

```sh
$ cd /usr/lib/x86_64-linux-gnu
$ ln -s libcurl.so.4 libcurl.so
$ ldconfig
```

On CentOS or Fedora:

```sh
$ cd /usr/lib64
$ ln -s libcurl.so.4 libcurl.so
$ ldconfig
```

vr009:
The problem appears when the module falls back to the system libcurl
(on relatively old tarantool versions) and when there is no dynamic library
named exactly as 'libcurl.so' (without the version suffix). This situation
usually occurs, when the module is installed from an RPM or Deb package
and the developer libcurl package is not installed.

Fixes #44

### DIFF
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -73,13 +73,15 @@ jobs:
           # uses the libcurl implementation provided by tarantool
           # executable and when the module falls back to the
           # system libcurl.
-          #
-          # TODO: Enable 1.10.7 jobs after resolving gh-44
-          # ('Unable to load libcurl.so'). Don't forget to exclude
-          # distribution versions, which are not supported by
-          # tarantool 1.10.7.
-          # - release/1.10.7
+          - release/1.10.7
           - live/2.9
+        exclude:
+          # Tarantool 1.10.7 release is not supported on
+          # debian:bullseye and fedora:34.
+          - platform: {os: debian, dist: bullseye}
+            tarantool: release/1.10.7
+          - platform: {os: fedora, dist: 34}
+            tarantool: release/1.10.7
 
     env:
       # Prevent packages like tzdata from asking configuration
--- a/smtp/smtpc.c
+++ b/smtp/smtpc.c
@@ -158,9 +158,9 @@ smtpc_init(void)
 
 	const char *libname;
 #ifdef __APPLE__
-	libname = "libcurl.dylib";
+	libname = "libcurl.4.dylib";
 #else
-	libname = "libcurl.so";
+	libname = "libcurl.so.4";
 #endif
 
 	/* Warn a user that we unable to use built-in libcurl. */


ligurio (changes requested):
Thanks for the patch!

Why all builds for macOS has failed?

```
-- Using bundled ZSTD
-- Could NOT find OpenSSL, try to set the path to OpenSSL root folder in the system variable OPENSSL_ROOT_DIR (missing: OPENSSL_INCLUDE_DIR) 
CMake Error at CMakeLists.txt:328 (message):
  Could NOT find OpenSSL development files (libssl-dev/openssl-devel package)
```

vr009:
> Why all builds for macOS has failed?
> 
> ```
> -- Using bundled ZSTD
> -- Could NOT find OpenSSL, try to set the path to OpenSSL root folder in the system variable OPENSSL_ROOT_DIR (missing: OPENSSL_INCLUDE_DIR) 
> CMake Error at CMakeLists.txt:328 (message):
>   Could NOT find OpenSSL development files (libssl-dev/openssl-devel package)
> ```

That problem relates to #49. I am currently fixing it.

ligurio:
> That problem relates to #49. I am currently fixing it.

Well, then we wait for a fix of #49.

The proposed change adds a condition specific for FreeBSD. It means that we need to build and test on FreeBSD in CI too.

Totktonada:
> The proposed change adds a condition specific for FreeBSD. It means that we need to build and test on FreeBSD in CI too.

Wow. I'm okay with just results on manual testing on FreeBSD shared here.

When we'll have an infrastructure for clean (created for a job) autoscaled self-hosted runners I'll be okay to add them into modules.

ligurio:
> When we'll have an infrastructure for clean (created for a job) autoscaled self-hosted runners I'll be okay to add them into modules.

It looks as a typical perfectionism: wait when our ideal CI infrastructure will work fine. Instead of using FreeBSD on cloud CI that just works (yep, I mean Cirrus CI).

Get me right: code that doesn't work is a dead code. FreeBSD-specific code is a dead code now.

Totktonada:
I like the idea with Cirrus CI. However for this particular change it will be okay for me if Slava will just share testing results on FreeBSD.

smtp/smtpc.c
```diff
@@ -158,9 +158,9 @@ smtpc_init(void)
 
 	const char *libname;
 #ifdef __APPLE__
-	libname = "libcurl.dylib";
+	libname = "libcurl.4.dylib";
```
Totktonada, line 161: Regarding Mac OS: the system library is versioned (has `.4`) and the library from homebrew as well, correct?

Totktonada:
Don't forget to enable the releavent testing job:

https://github.com/tarantool/smtp/blob/e17aad31f8caf1143374abaa0216d1651272afac/.github/workflows/packaging.yml#L77-L81

Totktonada:
The idea of the fix looks is okay, but I would highlight an alternative just in case.

Now we do the following:

* Try tarantool's built-in libcurl.
* Try `libcurl.so` / `libcurl.dylib`.
* Fail if no attempts were successful.

The safe way to fix the problem would be change the algorithm to the following:

* Try tarantool's built-in libcurl.
* Try `libcurl.so.4` / `libcurl.4.dylib`.
* Try `libcurl.so` / `libcurl.dylib`.
* Fail if no attempts were successful.

Is it possible that we'll have non-versioned library and will have no versioned one? I doubt about this. Disregard so. However we can return here if we'll see an actual problem.

smtp/smtpc.c
vr009: Yes, that's right.

vr009:
> Don't forget to enable the releavent testing job:
> 
> https://github.com/tarantool/smtp/blob/e17aad31f8caf1143374abaa0216d1651272afac/.github/workflows/packaging.yml#L77-L81

Sure I will enable this jobs, but I have faced a problem which is described in related patch #52. After this one will be solved I will check all other jobs for passing.

vr009:
> Is it possible that we'll have non-versioned library and will have no versioned one? I doubt about this. Disregard so. However we can return here if we'll see an actual problem.

Okay, I'm not sure that it is possible to have a non-versioned library without the versioned one.

smtp/smtpc.c
Totktonada: Okay so.

Totktonada:
Let's rebase after PR #52 and don't forget to uncomment related CI rules (see above).

vr009:
I have tried this changes on FreeBSD 12.2 and I had just one case when this module tries to fall back to system curl. That occurs with tarantool releases 1.10.x. The output was always like:
`root@freebsd:~/smtp # /root/tarantool/build/src/tarantool`
`Tarantool 1.10.7-0-g554d439a5`
`type 'help' for interactive help`
`tarantool> smtp = require('smtp').new()`
`Unable to load symbol curl_easy_cleanup from [tarantool]`
`Attempt to fallback to libcurl.so.4`
`Successfully loaded libcurl.so.4`

This fall back is also succesful with non-versioned libcurl, so we have both working options.

.github/workflows/packaging.yml
```diff
@@ -73,13 +73,13 @@ jobs:
...
-          # distribution versions, which are not supported by
-          # tarantool 1.10.7.
-          # - release/1.10.7
+          - release/1.10.7
           - live/2.9
+        exclude:
+          - platform: {os: debian, dist: bullseye}
+            tarantool: release/1.10.7
+          - platform: {os: fedora, dist: 34}
+            tarantool: release/1.10.7
```
Totktonada: Please, leave a comment with the reason of excluding. It may be not so obvious.

Totktonada (approved):
LGTM.

.github/workflows/packaging.yml
vr009: Added.

ligurio (approved):
LGTM

PR #52, opened 2021-12-19
github-ci: could not find openssl while building

### LINKED ISSUE #49: ci: 'Could NOT find OpenSSL' fails

vr009:
OSX workflows use brew for install openssl. Since brew
updated the openssl formula, it is needed to install
openssl@1.1 explicitly for successful build of Tarantool.

After the problem with openssl was solved, I have faced a problem which seems to be related to [#6576](https://github.com/tarantool/tarantool/issues/6576). 
This one seems to be fixed with this [commit](https://github.com/tarantool/tarantool/commit/d83e1901dc885234fe59eb326748c4d71e106c71). As I see, this patch have not reached Tarantool minor versions which are represented in `matrix` for workflow jobs, so all builds of unpatched versions are failed with this problem.

Fixes #49

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -198,7 +198,7 @@ jobs:
         if: matrix.tarantool != 'brew' && matrix.tarantool != 'master'
 
       - name: Install tarantool build dependencies
-        run: brew install autoconf automake libtool
+        run: brew install autoconf automake libtool openssl@1.1
         if: matrix.tarantool != 'brew' && steps.cache.outputs.cache-hit != 'true'
 
       - name: Clone tarantool ${{ env.T_VERSION }}
@@ -223,7 +223,11 @@ jobs:
           # There are tarantool releases on which AppleClang
           # complains about the problem that was fixed later in
           # https://github.com/tarantool/tarantool/commit/7e8688ff8885cc7813d12225e03694eb8886de29
-          cmake . -DCMAKE_BUILD_TYPE=RelWithDebInfo
+          #
+          # Set OpenSSL root directory for linking tarantool with OpenSSl of version 1.1
+          # This is related to #49. There are too much deprecations which affect the build and tests.
+          # Must be revisited after fixing https://github.com/tarantool/tarantool/issues/6477
+          cmake . -DCMAKE_BUILD_TYPE=RelWithDebInfo -DOPENSSL_ROOT_DIR=/usr/local/opt/openssl@1.1 -DOPENSSL_LIBRARIES=/usr/local/opt/openssl@1.1/lib
 
           # {{{ Workaround Mac OS build failure (gh-6076)
           #
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -214,6 +214,13 @@ jobs:
           fetch-depth: 0
         if: matrix.tarantool != 'brew' && steps.cache.outputs.cache-hit != 'true'
 
+      - name: Patching tarantool for successful build
+        run: |
+          # These steps fix the problem with tarantool build described in
+          # https://github.com/tarantool/tarantool/issues/6576
+          git show 11e87877df9001a4972019328592d79d55d1bb01 | patch -p1
+        if: matrix.tarantool != 'brew' && steps.cache.outputs.cache-hit != 'true'
+
       - name: Build tarantool ${{ env.T_VERSION }} from sources
         run: |
           mkdir "${T_DESTDIR}"
@@ -224,7 +231,7 @@ jobs:
           # complains about the problem that was fixed later in
           # https://github.com/tarantool/tarantool/commit/7e8688ff8885cc7813d12225e03694eb8886de29
           #
-          # Set OpenSSL root directory for linking tarantool with OpenSSl of version 1.1
+          # Set OpenSSL root directory for linking tarantool with OpenSSL of version 1.1
           # This is related to #49. There are too much deprecations which affect the build and tests.
           # Must be revisited after fixing https://github.com/tarantool/tarantool/issues/6477
           cmake . -DCMAKE_BUILD_TYPE=RelWithDebInfo -DOPENSSL_ROOT_DIR=/usr/local/opt/openssl@1.1 -DOPENSSL_LIBRARIES=/usr/local/opt/openssl@1.1/lib


.github/workflows/testing.yml
```diff
@@ -198,7 +198,14 @@ jobs:
...
       - name: Install tarantool build dependencies
-        run: brew install autoconf automake libtool
+        run: |
+          brew install autoconf automake libtool openssl@1.1
+          brew link --force openssl@1.1
+          export PATH="/usr/local/opt/openssl@1.1/bin:$PATH"
+          export LDFLAGS="-L/usr/local/opt/openssl@1.1/lib"
+          export CPPFLAGS="-I/usr/local/opt/openssl@1.1/include"
+          export PKG_CONFIG_PATH="/usr/local/opt/openssl@1.1/lib/pkgconfig"
+          export OPENSSL_ROOT_DIR="/usr/local/opt/openssl@1.1"
```
Totktonada: AFAIU, those exports are not seen on a next step. If they're actually necessary, use [GITHUB_ENV](https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#setting-an-environment-variable).

.github/workflows/testing.yml
```diff
@@ -198,7 +198,14 @@ jobs:
         if: matrix.tarantool != 'brew' && matrix.tarantool != 'master'
 
       - name: Install tarantool build dependencies
-        run: brew install autoconf automake libtool
+        run: |
+          brew install autoconf automake libtool openssl@1.1
+          brew link --force openssl@1.1
```
Totktonada: Whether we need this `brew link` command or just cmake options below are sufficient?

Totktonada:
> After the problem with openssl was solved, I have faced a problem which seems to be related to #6576.

Let's fix it as well in a second commit within this PR.

.github/workflows/testing.yml
vr009: CMake options seem to be sufficient here. Fixed.

.github/workflows/testing.yml
vr009: Sure, these exports are not necessary. I export all needed flags with calling CMake below. Fixed.

.github/workflows/testing.yml
```diff
@@ -214,6 +214,22 @@ jobs:
...
+        if: contains(matrix.tarantool, '1.10.') && steps.cache.outputs.cache-hit != 'true'
+
+      - name: Patching tarantool 2.4.x 2.5.x 2.6.x for succesful build
+        run: |
+            # These steps fix the problem with tarantool build described in #6576
+            touch temp
+            echo '#include <stdalign.h>' > temp
+            cat ${T_SRCDIR}/src/trivia/util.h >> temp
+            cat temp > ${T_SRCDIR}/src/trivia/util.h
+        if: matrix.runs-on == 'macos-11.0' && (contains(matrix.tarantool, '2.5.') || contains(matrix.tarantool, '2.4.') || contains(matrix.tarantool, '2.6.')) && steps.cache.outputs.cache-hit != 'true'
```
Totktonada:
```sh
git show 11e87877df9001a4972019328592d79d55d1bb01 | patch -p1
```

It succeeds on 1.10.7 and 2.4.1 for me. It seems you had some problem with just applying a patch. How can I reproduce it.

.github/workflows/testing.yml
```diff
@@ -214,6 +214,22 @@ jobs:
           fetch-depth: 0
         if: matrix.tarantool != 'brew' && steps.cache.outputs.cache-hit != 'true'
 
+      - name: Patching tarantool 1.10.x for succesful build
+        run: |
+          # These steps fix the problem with tarantool build described in #6576
```
Totktonada: Nit: I suggest to place a full link to the issue, because it is another repository.

.github/workflows/testing.yml
```diff
@@ -214,6 +214,22 @@ jobs:
           fetch-depth: 0
         if: matrix.tarantool != 'brew' && steps.cache.outputs.cache-hit != 'true'
 
+      - name: Patching tarantool 1.10.x for succesful build
```
Totktonada: Typo: succesful -> successful.

.github/workflows/testing.yml
```diff
@@ -198,7 +198,7 @@ jobs:
         if: matrix.tarantool != 'brew' && matrix.tarantool != 'master'
```
Totktonada, line 198: Both commits are marked as fixing the issue. Technically it is incorrect. Use either `Part of #xxx` + `Fixes #xxx` or `Fixes #xxx` + `Follows up #xxx` depending on which problem is stated in the issue.

.github/workflows/testing.yml
Totktonada: BTW, I would use the master commit (`11e87877df9001a4972019328592d79d55d1bb01`) instead of the cherry-pick to 1.10 (`d83e1901dc885234fe59eb326748c4d71e106c71`), but that's purely stylistic suggestion.

.github/workflows/testing.yml
```diff
@@ -198,7 +198,7 @@ jobs:
...
 
       - name: Install tarantool build dependencies
-        run: brew install autoconf automake libtool openssl@1.1
+        run: brew install autoconf automake libtool openssl@1.1 git
         if: matrix.tarantool != 'brew' && steps.cache.outputs.cache-hit != 'true'
```
Totktonada: Is it possible that we have no `git` installed on a GitHub Actions runner?

I created the following workflow in a testing repository:

```yml
name: Git on Mac OS

on: [push]

jobs:
  git_on_mac_os:
    strategy:
      matrix:
        runs-on:
          - macos-10.15
          - macos-11.0
    runs-on: ${{ matrix.runs-on }}
    steps:
      - run: git --version
```

And got the following:

* macos-10.15: git version 2.34.1.
* macos-11.0: git version 2.34.1.

.github/workflows/testing.yml
```diff
@@ -198,7 +198,7 @@ jobs:
         if: matrix.tarantool != 'brew' && matrix.tarantool != 'master'
```
Totktonada, line 199: Tarantool's issue in the commit message is not linked by just `#xxxx`, because it is in another repository. I would give a full link instead.

Totktonada: And, hm, title of the second commit is not relevant to what you're actually fixed there: it has nothing about openssl.

.github/workflows/testing.yml
```diff
@@ -223,7 +223,7 @@ jobs:
           # There are tarantool releases on which AppleClang
           # complains about the problem that was fixed later in
           # https://github.com/tarantool/tarantool/commit/7e8688ff8885cc7813d12225e03694eb8886de29
-          cmake . -DCMAKE_BUILD_TYPE=RelWithDebInfo
+          cmake . -DCMAKE_BUILD_TYPE=RelWithDebInfo -DOPENSSL_ROOT_DIR=/usr/local/opt/openssl@1.1 -DOPENSSL_LIBRARIES=/usr/local/opt/openssl@1.1/lib
```
Totktonada, line 237: I would leave a comment why explicit options related to openssl is passed there: just like build type is commented above. You can mention that this should be revisited after fixing of https://github.com/tarantool/tarantool/issues/6477.

Totktonada:
It seems, the fail on Fedora 34 is because we install only `tarantool` package and not `tarantool-dev` (Debian based) / `tarantool-devel` (RHEL based). It is irrelevant here, so I'll ignore this fail in this pull request.

.github/workflows/testing.yml
vr009: Thank you, this command applies the patch better. These steps are replaced in the latests update.

.github/workflows/testing.yml
vr009: Placed.

.github/workflows/testing.yml
vr009: Fixed.

.github/workflows/testing.yml
vr009: Fixed.

.github/workflows/testing.yml
vr009: Fixed all.

.github/workflows/testing.yml
vr009: Sure, the comment is left there in the latests update.

.github/workflows/testing.yml
vr009: I really shouldn't install it in runner. That was rather a little insurance. Now fixed.

.github/workflows/testing.yml
```diff
@@ -223,7 +223,11 @@ jobs:
           # There are tarantool releases on which AppleClang
           # complains about the problem that was fixed later in
           # https://github.com/tarantool/tarantool/commit/7e8688ff8885cc7813d12225e03694eb8886de29
-          cmake . -DCMAKE_BUILD_TYPE=RelWithDebInfo
+          #
+          # Set OpenSSL root directory for linking tarantool with OpenSSl of version 1.1
```
Totktonada: Typo: OpenSSl -> OpenSSL.

Totktonada (approved):
LGTM. Please, finish review with Sergey.

Some nits. I would make the second commit title/message more specific:

- 'patching tarantool for successful build' is a kind of 'fix some problem'. We can say at least 'fix tarantool build on Mac OS'. BTW, the rule of thumb is to use the imperative mood in a commit title.
- There is no information about the problem in the commit message. So even if I'm aware of the fix in tarantool, I can't deduce from the message what is the particular fix you say about (without going over links). Say, 'the build problem with `alignas` macro on Mac OS described in <...>' is just several more words, but allows to spot everything just from the commit message (if you looked at this before).

.github/workflows/testing.yml
vr009: Fixed.

vr009:
> Some nits. I would make the second commit title/message more specific:

Done.

Totktonada:
> It seems, the fail on Fedora 34 is because we install only `tarantool` package and not `tarantool-dev` (Debian based) / `tarantool-devel` (RHEL based). It is irrelevant here, so I'll ignore this fail in this pull request.

It occurs to be not so easy. I filed https://github.com/tarantool/smtp/issues/54 to hold symptoms of the problem. Let's wait for a fix on the RPM repositories infrastructure side.

vr009:
> It occurs to be not so easy. I filed #54 to hold symptoms of the problem. Let's wait for a fix on the RPM repositories infrastructure side.

The problem seems to be fixed.

Totktonada (approved):

PR #53, opened 2021-12-19
client: handling SMTP errors from server

### LINKED ISSUE #13: Any 5xx error from SMTP server becomes a strange internal error

vr009:
In case server has some settings, which don't let some
types of SMTP request to be handled, SMTP client may receive
5xx, 4xx, 3xx codes and other malformed answers. This patch fixes 
the problem, when on the client side this SMTP codes are not handled normally.
Instead of runtime error and unusual message that there is
some connection error.

Fixes #13

### DIFF
--- a/smtp/smtpc.c
+++ b/smtp/smtpc.c
@@ -241,6 +241,14 @@ smtpc_request_new(struct smtpc_env *env, const char *url, const char *from)
 		return NULL;
 	}
 	memset(req, 0, sizeof(*req));
+
+	req->error_buf = calloc(CURL_ERROR_SIZE, sizeof(char));
+	if (req->error_buf == NULL) {
+		free(req);
+		box_error_set(__FILE__, __LINE__, ER_MEMORY_ISSUE,
+					  "Can't alloc error message buffer");
+		return NULL;
+	}
 	req->env = env;
 
 	req->easy = curl_easy_init();
@@ -252,6 +260,7 @@ smtpc_request_new(struct smtpc_env *env, const char *url, const char *from)
 	}
 	curl_easy_setopt(req->easy, CURLOPT_URL, url);
 	curl_easy_setopt(req->easy, CURLOPT_MAIL_FROM, from);
+	curl_easy_setopt(req->easy, CURLOPT_ERRORBUFFER, req->error_buf);
 
 	return req;
 }
@@ -270,6 +279,7 @@ smtpc_request_delete(struct smtpc_request *req)
 	if (req->easy != NULL)
 		coio_call(smtpc_task_delete, req);
 	free(req->body);
+	free(req->error_buf);
 	if (req->recipients)
 		curl_slist_free_all(req->recipients);
 
@@ -440,10 +450,26 @@ smtpc_execute(struct smtpc_request *req, double timeout)
 		++env->stat.failed_requests;
 		smtpc_request_delete(req);
 		return -1;
+	case CURLE_SEND_ERROR:
+	case CURLE_RECV_ERROR:
+		curl_easy_getinfo(req->easy, CURLINFO_RESPONSE_CODE, &longval);
+		req->status = (int) longval;
+		/*
+		 * When client can not read bytes from another side that
+		 * leads to a message about failed response reading but
+		 * response code stays 250, that is not correct. See
+		 * https://github.com/curl/curl/issues/8356
+		 */
+		if (strcmp(req->error_buf, "response reading failed") == 0) {
+			req->status = -1;
+		}
+		req->reason = req->error_buf;
+		++env->stat.failed_requests;
+		break;
 	default: {
 		char error_msg[256];
 		curl_easy_getinfo(req->easy, CURLINFO_OS_ERRNO, &longval);
-		snprintf(error_msg, sizeof(error_msg), "SMTP error %i (os errno %li)", req->code, longval);
+		snprintf(error_msg, sizeof(error_msg), "CURL error %i (os errno %li)", req->code, longval);
 		box_error_set(__FILE__, __LINE__, ER_UNKNOWN, error_msg);
 		++env->stat.failed_requests;
 		smtpc_request_delete(req);
--- a/smtp/smtpc.h
+++ b/smtp/smtpc.h
@@ -112,10 +112,27 @@ struct smtpc_request {
 	int body_size;
 	/** Buffer read position. */
 	char *body_rpos;
-	/** SMTP status code. */
+	/**
+	 * SMTP status code.
+	 * It takes the value of -1 if there is some problem,
+	 * which is not related to SMTP, like connection error.
+	 */
 	int status;
-	/** Error message. */
+	/**
+	 * Error message.
+	 * It is a string to report details of an error to a user.
+	 * Does not require freeing.
+	 * It is never NULL if smtpc_execute() returns zero.
+	 */
 	const char *reason;
+	/**
+	 * Error buffer for receiving messages.
+	 * It should exist during a request lifetime and
+	 * must be freed at freeing the request structure.
+	 * This field is not for reading directly, cause
+	 * reason field points to it, when appropriate.
+	 */
+	char *error_buf;
 };
 
 /**
--- a/test/smtp.test.lua
+++ b/test/smtp.test.lua
@@ -10,6 +10,21 @@ local os = require('os')
 test:plan(1)
 mails = fiber.channel(100)
 
+function write_reply_code(s, l)
+    if l:find('3xx') then
+        s:write('354 Start mail input\r\n')
+    elseif l:find('4xx') then
+        s:write('421 Service not available, closing transmission channel\r\n')
+    elseif l:find('5xx') then
+        s:write('510 Bad email address\r\n')
+    elseif l:find('breakconnect') then
+        return -1
+    else
+        s:write('250 OK\r\n')
+    end
+    return 1
+end
+
 local function smtp_h(s)
     s:write('220 localhost ESMTP Tarantool\r\n')
     local l
@@ -26,10 +41,14 @@ local function smtp_h(s)
             s:write('250 HELP\r\n')
         elseif l:find('MAIL FROM:') then
             mail.from = l:sub(11):sub(1, -3)
-            s:write('250 OK\r\n')
+            if write_reply_code(s, l) == -1 then
+                return
+            end
         elseif l:find('RCPT TO:') then
             mail.rcpt[#mail.rcpt + 1] = l:sub(9):sub(1, -3)
-            s:write('250 OK\r\n')
+            if write_reply_code(s, l) == -1 then
+                return
+            end
         elseif l == 'DATA\r\n' then
             s:write('354 Enter message, ending with "." on a line by itself\r\n')
             while true do
@@ -56,7 +75,7 @@ local server = socket.tcp_server('127.0.0.1', 0, smtp_h)
 local addr = 'smtp://127.0.0.1:' .. server:name().port
 
 test:test("smtp.client", function(test)
-    test:plan(10)
+    test:plan(26)
     local r
     local m
 
@@ -146,6 +165,54 @@ test:test("smtp.client", function(test)
                   m.text,
                   "Subject: =%?utf%-8%?b%?YWJjZGVmZ2hpamvRj2xtbm9wcXJzdHV2d3h5eg==%?=", ""))
     test:is(subj, 1, 'subject codes >127')
-end)
 
+    r = client:request(addr, '3xx@tarantool.org',
+                       'receiver@tarantool.org',
+                       'mail.body')
+    test:is(r.reason, 'MAIL failed: 354', 'errors 3xx')
+    test:is(r.status, 354, 'expected code')
+
+    r = client:request(addr, '4xx@tarantool.org',
+                       'receiver@tarantool.org',
+                       'mail.body')
+    test:is(r.reason, 'MAIL failed: 421', 'service unavailable')
+    test:is(r.status, 421, 'expected code')
+
+    r = client:request(addr, '5xx@tarantool.org',
+                       'receiver@tarantool.org',
+                       'mail.body')
+    test:is(r.reason, 'MAIL failed: 510', 'unexisting recipient')
+    test:is(r.status, 510, 'expected code')
+
+    r = client:request(addr, 'breakconnect@tarantool.org',
+                       'receiver@tarantool.org',
+                       'mail.body')
+    test:is(r.reason, 'response reading failed', 'unexpected response code')
+    test:is(r.status, -1, 'expected code')
+
+    r = client:request(addr, 'sender@tarantool.org',
+                       '3xx@tarantool.org',
+                       'mail.body')
+    test:is(r.reason, 'RCPT failed: 354', 'errors 3xx')
+    test:is(r.status, 354, 'expected code')
+
+    r = client:request(addr, 'sender@tarantool.org',
+                       '4xx@tarantool.org',
+                       'mail.body')
+    test:is(r.reason, 'RCPT failed: 421', 'service unavailable')
+    test:is(r.status, 421, 'expected code')
+
+    r = client:request(addr, 'sender@tarantool.org',
+                       '5xx@tarantool.org',
+                       'mail.body')
+    test:is(r.reason, 'RCPT failed: 510', 'unexisting recipient')
+    test:is(r.status, 510, 'expected code')
+
+    r = client:request(addr, 'sender@tarantool.org',
+                       'breakconnect@tarantool.org',
+                       'mail.body')
+    test:is(r.reason, 'response reading failed', 'unexpected response code')
+    test:is(r.status, -1, 'expected code')
+
+end)
 os.exit(test:check() == true and 0 or -1)
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -194,7 +194,10 @@ jobs:
           #
           # v3 is to re-verify all Mac OS builds after fix for the
           # gh-6076 problem (see below).
-          key: ${{ matrix.runs-on }}-${{ matrix.tarantool }}-v3
+          #
+          # v4 added due to inability to clear the cache after v3 prefix.
+          # See https://github.com/github/docs/issues/14145
+          key: ${{ matrix.runs-on }}-${{ matrix.tarantool }}-v4
         if: matrix.tarantool != 'brew' && matrix.tarantool != 'master'
 
       - name: Install tarantool build dependencies


smtp/smtpc.c
```diff
@@ -241,6 +241,14 @@ smtpc_request_new(struct smtpc_env *env, const char *url, const char *from)
 		return NULL;
 	}
 	memset(req, 0, sizeof(*req));
+	req->smtp_error = (char *) malloc(CURL_ERROR_SIZE + sizeof(" (SMTP error)"));
+	if (req->smtp_error == NULL) {
+		free(req);
+		box_error_set(__FILE__, __LINE__, ER_MEMORY_ISSUE,
+		              "Can't alloc curl buffer for error message");
```
ligurio: Hm, seems nothing specific for curl here. Why it is a "curl buffer"?

smtp/smtpc.c
```diff
@@ -440,10 +452,21 @@ smtpc_execute(struct smtpc_request *req, double timeout)
...
+		}
+		strcat(req->smtp_error, " (SMTP error)");
+		req->reason = req->smtp_error;
+		++env->stat.failed_requests;
+		break;
 	default: {
 		char error_msg[256];
 		curl_easy_getinfo(req->easy, CURLINFO_OS_ERRNO, &longval);
-		snprintf(error_msg, sizeof(error_msg), "SMTP error %i (os errno %li)", req->code, longval);
+		snprintf(error_msg, sizeof(error_msg), "CURL error %i (os errno %li)", req->code, longval);
```
ligurio, line 472: Not sure regarding this change. Tarantool user uses SMTP module, not a curl, he knows nothing what SMTP module uses under the hood.

smtp/smtpc.h
```diff
@@ -116,6 +116,7 @@ struct smtpc_request {
 	int status;
 	/** Error message. */
 	const char *reason;
+	char *smtp_error;
```
ligurio: Perhaps you can name as 'error' because structure used for smtp request already contains the prefix 'smtp' and we can omit prefix 'smtp' for variable with error. I will not insist.

test/smtp.test.lua
```diff
@@ -54,9 +76,9 @@ end
 
 local server = socket.tcp_server('127.0.0.1', 0, smtp_h)
 local addr = 'smtp://127.0.0.1:' .. server:name().port
-
+print(addr)
```
ligurio: Is it forgotten debug print?

ligurio (changes requested):
Thanks for the patch!

smtp/smtpc.h
vr009: Fixed.

smtp/smtpc.c
vr009: You are right. Fixed.

test/smtp.test.lua
vr009: Fixed :)

smtp/smtpc.c
vr009: The codes which are represented in this line have nothing to do with SMTP codes. So that must be either some CURL error code with its value or I would just print OS error code here.

test/smtp.test.lua
```diff
@@ -146,6 +168,48 @@ test:test("smtp.client", function(test)
...
+    test:is(r.reason, 'RCPT failed: 421 (SMTP error)', 'service unavailable')
+
+    r = client:request(addr, 'sender@tarantool.org',
+            '5xx@tarantool.org',
+            'mail.body')
+    test:is(r.reason, 'RCPT failed: 510 (SMTP error)', 'unexisting recipient')
+
+    r = client:request(addr, 'sender@tarantool.org',
+            '10xx@tarantool.org',
+            'mail.body')
```
ligurio: Could you indent arguments on newlines?
```lua
    r = client:request(addr, 'sender@tarantool.org',
                       '5xx@tarantool.org',
                       'mail.body')
    test:is(r.reason, 'RCPT failed: 510 (SMTP error)', 'unexisting recipient')
```

smtp/smtpc.h
```diff
@@ -116,6 +116,7 @@ struct smtpc_request {
 	int status;
 	/** Error message. */
 	const char *reason;
+	char *error;
```
ligurio: Why we cannot reuse `reason` for our purposes?

ligurio (changes requested):

smtp/smtpc.h
vr009: The signautre of function `CURLcode curl_easy_setopt(CURL *handle, CURLOPT_ERRORBUFFER, char *buf);` accepts a non-constant char buffer. The reference is [here](https://curl.se/libcurl/c/CURLOPT_ERRORBUFFER.html).

test/smtp.test.lua
vr009: Done.

ligurio: Sorry, but indentation is still not perfect.

Screenshot from 2022-01-11 09-44-37

I have fixed it in a separate patch and put it on top of your branch:

```diff
--- a/test/smtp.test.lua
+++ b/test/smtp.test.lua
@@ -170,43 +170,43 @@ test:test("smtp.client", function(test)
     test:is(subj, 1, 'subject codes >127')
 
     r = client:request(addr, '3xx@tarantool.org',
-                        'receiver@tarantool.org',
-                        'mail.body')
+                       'receiver@tarantool.org',
+                       'mail.body')
     test:is(r.reason, 'MAIL failed: 354 (SMTP error)', 'Errors 3xx')
 
     r = client:request(addr, '4xx@tarantool.org',
-                        'receiver@tarantool.org',
-                        'mail.body')
+                       'receiver@tarantool.org',
+                       'mail.body')
     test:is(r.reason, 'MAIL failed: 451 (SMTP error)', 'service unavailable')
 
     r = client:request(addr, '5xx@tarantool.org',
-                        'receiver@tarantool.org',
-                        'mail.body')
+                       'receiver@tarantool.org',
+                       'mail.body')
     test:is(r.reason, 'MAIL failed: 510 (SMTP error)', 'unexisting recipient')
 
     r = client:request(addr, '10xx@tarantool.org',
-                        'receiver@tarantool.org',
-                        'mail.body')
+                       'receiver@tarantool.org',
+                       'mail.body')
     test:is(r.reason, 'response reading failed (SMTP error)', 'unexisting recipient')
 
     r = client:request(addr, 'sender@tarantool.org',
-                        '3xx@tarantool.org',
-                        'mail.body')
+                       '3xx@tarantool.org',
+                       'mail.body')
     test:is(r.reason, 'RCPT failed: 354 (SMTP error)', 'Errors 3xx')
 
     r = client:request(addr, 'sender@tarantool.org',
-                        '4xx@tarantool.org',
-                        'mail.body')
+                       '4xx@tarantool.org',
+                       'mail.body')
     test:is(r.reason, 'RCPT failed: 421 (SMTP error)', 'service unavailable')
 
     r = client:request(addr, 'sender@tarantool.org',
-                        '5xx@tarantool.org',
-                        'mail.body')
+                       '5xx@tarantool.org',
+                       'mail.body')
     test:is(r.reason, 'RCPT failed: 510 (SMTP error)', 'unexisting recipient')
 
     r = client:request(addr, 'sender@tarantool.org',
-                        '10xx@tarantool.org',
-                        'mail.body')
+                       '10xx@tarantool.org',
+                       'mail.body')
     test:is(r.reason, 'response reading failed (SMTP error)', 'unexisting recipient')
 
 end)
```

ligurio (approved):
Thanks for the patch!

LGTM after fixing indentation in test.

smtp/smtpc.c
```diff
@@ -241,6 +241,14 @@ smtpc_request_new(struct smtpc_env *env, const char *url, const char *from)
 		return NULL;
 	}
 	memset(req, 0, sizeof(*req));
+	req->error = (char *) malloc(CURL_ERROR_SIZE + sizeof(" (SMTP error)"));
```
olegrok: C doesn't require explicit casts. So it could be just `req->error = malloc(...)`.

BTW if you plan to use `memset(req, 0, sizeof(*req->error))` why don't you use calloc.

smtp/smtpc.c
```diff
@@ -241,6 +241,14 @@ smtpc_request_new(struct smtpc_env *env, const char *url, const char *from)
...
+	if (req->error == NULL) {
+		free(req);
+		box_error_set(__FILE__, __LINE__, ER_MEMORY_ISSUE,
+					  "Can't alloc error message buffer");
+		return NULL;
+	}
+	memset(req, 0, sizeof(*req->error));
```
olegrok: I guess result will be wrong. `sizeof(*req->error)` == `sizeof(char)` (not whole string).

olegrok (changes requested):

test/smtp.test.lua
vr009: Thank you!

smtp/smtpc.c
vr009: Fixed.

smtp/smtpc.c
vr009: Thank you for idea, fixed.

vr009: Thank you for idea, fixed.

olegrok (approved):

test/smtp.test.lua
Totktonada: Seems resolved, so I'll close the discussion.

smtp/smtpc.c
```diff
@@ -380,6 +391,7 @@ smtpc_execute(struct smtpc_request *req, double timeout)
 {
 	struct smtpc_env *env = req->env;
 
+	curl_easy_setopt(req->easy, CURLOPT_ERRORBUFFER, req->error);
```
Totktonada: We already did it in `smtpc_request_new()`, no?

smtp/smtpc.c
```diff
@@ -270,6 +279,8 @@ smtpc_request_delete(struct smtpc_request *req)
 	if (req->easy != NULL)
 		coio_call(smtpc_task_delete, req);
 	free(req->body);
+	if (req->error != NULL)
+		free(req->error);
```
Totktonada: AFAIU, it is never `NULL`.

test/smtp.test.lua
```diff
@@ -143,9 +165,50 @@ test:test("smtp.client", function(test)
...
+
+    r = client:request(addr, 'sender@tarantool.org',
+                       '5xx@tarantool.org',
+                       'mail.body')
+    test:is(r.reason, 'RCPT failed: 510 (SMTP error)', 'unexisting recipient')
+
+    r = client:request(addr, 'sender@tarantool.org',
+                       '10xx@tarantool.org',
+                       'mail.body')
+    test:is(r.reason, 'response reading failed (SMTP error)', 'unexisting recipient')
```
Totktonada: The test case description says 'unexisting recipient', but it is about unexpected response code.

test/smtp.test.lua
```diff
@@ -143,9 +165,50 @@ test:test("smtp.client", function(test)
 
     m = mails:get()
     subj = select(2, string.gsub(
-                  m.text,
-                  "Subject: =%?utf%-8%?b%?YWJjZGVmZ2hpamvRj2xtbm9wcXJzdHV2d3h5eg==%?=", ""))
+                        m.text,
+                "Subject: =%?utf%-8%?b%?YWJjZGVmZ2hpamvRj2xtbm9wcXJzdHV2d3h5eg==%?=", ""))
```
Totktonada: The indent here looks strange. And you didn't change this code aside of the indent, so you can revert the change.

test/smtp.test.lua
```diff
@@ -143,9 +165,50 @@ test:test("smtp.client", function(test)
...
     test:is(subj, 1, 'subject codes >127')
+
+    r = client:request(addr, '3xx@tarantool.org',
+                       'receiver@tarantool.org',
+                       'mail.body')
+    test:is(r.reason, 'MAIL failed: 354 (SMTP error)', 'Errors 3xx')
```
Totktonada: How about verifying `r.status` as well? I'm interested here not even in a regression test, but rather illustration of the idea what to expect in the field.

test/smtp.test.lua
```diff
@@ -26,10 +26,32 @@ local function smtp_h(s)
...
+            elseif l:find('4xx') then
+                s:write('421 Service not available, closing transmission channel\r\n')
+            elseif l:find('5xx') then
+                s:write('510 Bad email address\r\n')
+            elseif l:find('10xx') then
+                s:write('1000 UNKNOWN ERROR\r\n')
+                return
+            else
+                s:write('250 OK\r\n')
+            end
```
Totktonada: Why we duplicated this code? Can we avoid the duplication somehow?

smtp/smtpc.c
```diff
@@ -440,10 +452,21 @@ smtpc_execute(struct smtpc_request *req, double timeout)
 		++env->stat.failed_requests;
 		smtpc_request_delete(req);
 		return -1;
+	case CURLE_SEND_ERROR:
+	case CURLE_RECV_ERROR:
+		curl_easy_getinfo(req->easy, CURLINFO_RESPONSE_CODE, &longval);
+		req->status = (int) longval;
+		if (strcmp(req->error, "response reading failed") == 0) {
+			req->status = -1;
+		}
```
Totktonada: This logic requires a clarification. At least it raises the following questions in my head:

1. `-1` is here, when..? Can we formalize the contract and place it to the structure description?
2. Why we handle one partucular case ('response reading failed') separately from others?
3. Why we do that using `strcmp()`, is there other way?

I would like to see those questions answered by the code (including comments where appropriate), not just discuss it and forget.

smtp/smtpc.c
```diff
@@ -440,10 +452,21 @@ smtpc_execute(struct smtpc_request *req, double timeout)
...
+		strcat(req->error, " (SMTP error)");
```
Totktonada: Is it really necessary? AFAIS, the error messages are quite descriptive itself. We can simplify the code a bit so.

smtp/smtpc.h
Totktonada: We have `reason` that is commented as 'error message' and `error` without any comment (is not it an 'error message'?). It does not look descriptive enough. Let's think about proper comments together.

First, what reader's questions the comments should answer?

1. How the field is intended to use, what to expect here in the general case, what are corner cases (say, a value at request failure).
2. How to manage memory pointed by pointers.

So we should highlight that the `reason` field is always filled in `smtpc_execute()` if it returns zero and is `NULL` otherwise. We should clarify that it is a string to report details of an error to a user. We should mention that it does not require freeing (points to a static buffer or to `error`).

The `error` is actually a temporary buffer that should exist during a request lifetime. It is not expected to be read directly, because `reason` points to it, when appropriate. It is always non-NULL and may contain zeros or a zero terminated error message depending on how the request is failed. It must be freed at freeing the request structure (but not before that to don't corrupt `reason`).

I would rename the `error` field: since it is just buffer, maybe `error_buf` will reflect its usage better?

Totktonada (commented):
Thanks for the patch! Please, consider my inline comments.

smtp/smtpc.c
vr009: You are right, fixed now.

smtp/smtpc.c
vr009: You are right, after allocation this buffer in `smtpc_request_new` we check if it is `NULL`, so that is imposiible.

smtp/smtpc.c
vr009: 1. First of all request->status must contain some SMTP code. But with this value I meaned that there is some situation when this code does not correspond to reality. For example the next point.
2. I have catched some specific case, when the client side cant get bytes from reposnse, in this case we get `CURLE_RECV_ERROR` and the message `response reading failed`. Without changing status code I receive 250. 
I am not sure that it is a correct behavior. So that looks like:
   status: 250
   reason: response reading failed
3. This way I don't see another workaround of this case.

smtp/smtpc.h
vr009: Thank you for explanation! I've renamed the field and added some comments there that explain it.

smtp/smtpc.c
vr009: Actually not. Fixed.

test/smtp.test.lua
vr009: Sure, fixed.

smtp/smtpc.c
vr009: The place in source code of CURL, where this message get collected to buffer is [here](https://github.com/curl/curl/blob/e7416cfd2bd58d256b8524f31ef22a43aa23a970/lib/pingpong.c#L342).

test/smtp.test.lua
vr009: That looks more like receiving malformed answer, cause, for example, `1000` status code is not typicall for SMTP that is why there are some troubles with handling this type of repsonses. For now this test is about another failre, so that description is changed, cause there is no handling of malformed answers yet. I'm working on it currently.

test/smtp.test.lua
vr009: Fixed.

test/smtp.test.lua
vr009: If we have some contract regarding this field, we could check this value according to it. So If it is SMTP status, we check SMTP status.

smtp/smtpc.c
vr009: Maybe I would prefer to handle this case as some CURL or OS error, so I would forward it further to default case [here](https://github.com/tarantool/smtp/blob/master/smtp/smtpc.c#L443). What do you think about it?

vr009:
I have investigated the reason of hanging up of a client after malformed response from server. It seems there is no normal workaround of this problem. It has to wait until the end of timeout. Also there is no way to add custom smtp codes which are not 3 digit numbers. The reason is [here](https://github.com/curl/curl/blob/7a92f865b43aac942ae5755a7a3a06775751fdea/lib/smtp.c#L212).

smtp/smtpc.h
```diff
@@ -112,10 +112,22 @@ struct smtpc_request {
 	int body_size;
 	/** Buffer read position. */
 	char *body_rpos;
-	/** SMTP status code. */
+	/** SMTP status code.
+	 * It takes the value of -1 if SMTP code doesn't match to error status.
+	 */
```
Totktonada: We use two styles of comments (for one-line and multi-line comments):

```c
/* Hello. */
```

```c
/*
 * Hello,
 * world.
 */
```

(Here the opening `/*` is `/**` for documentation comments.)

If you're curious, Linus about comments style: https://lkml.org/lkml/2016/7/8/625

smtp/smtpc.c
```diff
@@ -440,10 +450,25 @@ smtpc_execute(struct smtpc_request *req, double timeout)
...
 		return -1;
+	case CURLE_SEND_ERROR:
+	case CURLE_RECV_ERROR:
+		curl_easy_getinfo(req->easy, CURLINFO_RESPONSE_CODE, &longval);
+		req->status = (int) longval;
+
+		/** When client can not read bytes from another side that
+		 * leads to a message about failed response reading but
+		 * response code stays 250, that is not correct.
+		 */
```
Totktonada: This is not a documentation comment, but rather a code comment (don't know how to name it properly, but I guess you spotted the difference). So it should start from `/*`.

test/smtp.test.lua
Totktonada: Without a contract it has no much sense. So I propose to deduce its meaning from the code, clarify in the structure field description (it is done, as I see) and add regression test cases (just verify the codes here).

smtp/smtpc.h
vr009: Thank you for explanation of comments style! But I see that other fields of this structrure are described with documentation comments. Is that okay, if those fields will have different types of comments?

smtp/smtpc.c
Totktonada: Thanks for the explanation!

It looks as a problem in libcurl: there is no reliable way to obtain an SMTP reply code (and be sure that it is actually sent by server in a response, not something curl-specific). It is especially strange to get 2xx reply code for failure at reading.

If you don't mind to send a bug report to curl (you need to create a pure C reproducer for that) and clarify here that you need to deliver SMTP reply code (and not something else) to a caller, it would be nice to link it from the comment in the code.

We had a problem of this kind with HTTP(S): a write error was reported instead of a more specific error. It was fixed in libcurl: https://github.com/curl/curl/issues/4310.

Regarding the `default` switch case: that's strange to have two error reporting formats. I think we should revisit it. Filed https://github.com/tarantool/smtp/issues/57.

smtp/smtpc.h
Totktonada: I'm specifically about formatting. For documentation comments it'll be

```c
/** Hello. */
```

and

```c
/**
 * Hello,
 * world.
 */
```

smtp/smtpc.h
Totktonada: Those comments resolve the main problem: not it is clear what are `reason` and `error_buf`.

Several nits:

> It takes the value of -1 if SMTP code doesn't match to error status.

I don't understand the 'code doesn't match to error status' phrase, to be honest.

There is no information that `reason` can be `NULL` (and when).

test/smtp.test.lua
```diff
@@ -10,6 +10,22 @@ local os = require('os')
 test:plan(1)
 mails = fiber.channel(100)
 
+function mail_sort(s, l)
```
Totktonada: Sort? Maybe, `write_reply_code()`?

Totktonada:
I finished the review iteration except the problem, where the client hangs. The latter is not clear for me, so I would prefer to discuss it voicely. We'll talk with Slava tomorrow and return here with updates.

test/smtp.test.lua
vr009: Yes, that sounds better. Fixed.

smtp/smtpc.c
vr009: Fixed.

smtp/smtpc.h
vr009: Fixed.

smtp/smtpc.h
vr009: Fixed.

test/smtp.test.lua
vr009: Added.

test/smtp.test.lua
```diff
@@ -146,6 +166,52 @@ test:test("smtp.client", function(test)
...
+    r = client:request(addr, 'sender@tarantool.org',
+                       '5xx@tarantool.org',
+                       'mail.body')
+    test:is(r.reason, 'RCPT failed: 510', 'unexisting recipient')
+    test:is(r.status, 510, 'Unexpected code')
+
+    r = client:request(addr, 'sender@tarantool.org',
+                       'breakconnect@tarantool.org',
+                       'mail.body')
+    test:is(r.reason, 'response reading failed', 'unexpected response code')
```
Totktonada: What is the idea behind duplicating those test cases with error injection using a sender address and a receiver one?

test/smtp.test.lua
```diff
@@ -146,6 +166,52 @@ test:test("smtp.client", function(test)
...
+    r = client:request(addr, '5xx@tarantool.org',
+                       'receiver@tarantool.org',
+                       'mail.body')
+    test:is(r.reason, 'MAIL failed: 510', 'unexisting recipient')
+    test:is(r.status, 510, 'Unexpected code')
+
+    r = client:request(addr, 'breakconnect@tarantool.org',
+                       'receiver@tarantool.org',
+                       'mail.body')
+    test:is(r.reason, 'response reading failed', 'unexpected response code')
```
Totktonada, line 190: Maybe test that we have `-1` is `r.status` here?

test/smtp.test.lua
```diff
@@ -10,6 +10,22 @@ local os = require('os')
...
+function write_reply_code(s, l)
+    if l:find('3xx') then
+        s:write('354 Start mail input\r\n')
+    elseif l:find('4xx') then
+        s:write('421 Service not available, closing transmission channel\r\n')
+    elseif l:find('5xx') then
+        s:write('510 Bad email address\r\n')
+    elseif l:find('breakconnect') then
+        s:write('1000 UNKNOWN ERROR\r\n')
+        return -1
```
Totktonada: The idea is to test an unexpected reply code or the connection break situation? Now it looks as a strange mix.

smtp/smtpc.h
Totktonada: I looked again and, it seems, I was not completely right. I'll cite the new part of your comment:

> \* It can be NULL in some exceptional situations,
> \* when there is some Curl internal memory issue
> \* or returned Curl code is unexpected.

But in those cases the request is deleted and accessing this field is illegal anyway. So we can state that it is defined and never NULL if `smtpc_execute()` returns zero.

vr009: Fixed.

test/smtp.test.lua
vr009: The idea is that there are two phases in SMTP connection, when we can get some SMTP error from server. If server doesn't want to or can't work with sender and if it doesn't want to or can't work with recipient. We should make sure that both cases can be handled.

test/smtp.test.lua
vr009: Added.

test/smtp.test.lua
vr009: Fixed.

smtp/smtpc.c
vr009: I'm working on it currently. I will keep you up to date about this.

smtp/smtpc.h
Totktonada: Thanks!

test/smtp.test.lua
Totktonada: Okay, thanks for the explanation.

test/smtp.test.lua
Totktonada: Why it wrotes 'Breaking connection\r\n' to the client? Shouldn't it just break the connection?

vr009: Yes sure, that sounds more correctly. Fixed.

Totktonada: I still see `s:write('Break the connection\r\n')` here. Are we understood each other?

vr009: Sorry, firstly not, but now i got it.

Totktonada: Looks ok now.

smtp/smtpc.c
vr009: Filed in https://github.com/curl/curl/issues/8356.

Totktonada:
CentOS 8 job fails with the following diagnostics:

> Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist

And the following steps give me the same:

```
$ docker run -it centos:8
# dnf makecache
```

So it looks more like an infrastructure problem. I would ignore it here.

Totktonada (approved):
LGTM.

PR #56, opened 2022-01-14
github-ci: fix tarantool build on MacOS

vr009:
Without this update the patch that fixes tarantool build
on Mac OS is not working. That problem firstly was not detected
cause of caching some CI-steps. The original problem described
in https://github.com/tarantool/tarantool/issues/6576.

Follows up #49

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -216,10 +216,11 @@ jobs:
 
       - name: Patching tarantool for successful build
         run: |
+          cd "${T_SRCDIR}"
           # These steps fix the problem with tarantool build described in
           # https://github.com/tarantool/tarantool/issues/6576
-          git show 11e87877df9001a4972019328592d79d55d1bb01 | patch -p1
-        if: matrix.tarantool != 'brew' && steps.cache.outputs.cache-hit != 'true'
+          git show 11e87877df9001a4972019328592d79d55d1bb01 | patch -p1 -f
+        if: matrix.tarantool != 'brew' && matrix.tarantool != 'master' && steps.cache.outputs.cache-hit != 'true'
 
       - name: Build tarantool ${{ env.T_VERSION }} from sources
         run: |


.github/workflows/testing.yml
```diff
@@ -216,13 +216,15 @@ jobs:
...
           # These steps fix the problem with tarantool build described in
           # https://github.com/tarantool/tarantool/issues/6576
-          git show 11e87877df9001a4972019328592d79d55d1bb01 | patch -p1
-        if: matrix.tarantool != 'brew' && steps.cache.outputs.cache-hit != 'true'
+          git show 11e87877df9001a4972019328592d79d55d1bb01 | patch -p1 -f
+        if: matrix.tarantool != 'brew' && matrix.tarantool != 'master' && steps.cache.outputs.cache-hit != 'true'
 
       - name: Build tarantool ${{ env.T_VERSION }} from sources
         run: |
+          rm -rf "${T_DESTDIR}"
```
Totktonada: I don't see how we can get this directory existing. Can you clarify, please?

vr009: I get the message in Ci that this directory already exists [here](https://github.com/tarantool/smtp/runs/4813668988?check_suite_focus=true). Command `rm -rf` doesn't return an error at all, so that will not affect workflow even if there is no directory there.

Totktonada: Here we got the cache restored despite the `steps.cache.outputs.cache-hit != 'true'` condition. To be honest, I don't understand how we can reach this situation: it looks as unreachable.

vr009: Me too, but you can see this situation [here](https://github.com/tarantool/smtp/runs/4813668988?check_suite_focus=true).

Totktonada: I see. Is it a reproducible situation? Can you debug it a bit (print the `steps.cache.outputs.cache-hit` value at least)? Also I temporary enabled [debug logs](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging) for the repository.

vr009: I don't think that it is a reproducable sitution. For now it works correctly as it has to. But the failed case had the next scenario. The worflow run started with one key, let it be `key-name`. But cache was reloaded from key `key-name-test`. At the same time we had two different keys, so the `cache.hit` is `false`, that is why all the steps were working. That leaded to all errors.

Totktonada: Thanks for the investigation! I see, the cache was restored from the key `macos-10.15-2.7.1-v3-test`, while the `key` field is `${{ matrix.runs-on }}-${{ matrix.tarantool }}-v3`. It seems, the documentation is not clear here.

Disabled debugging logs.

Totktonada (approved):
LGTM.

.github/workflows/testing.yml
Totktonada: I created an issue aganst GitHub documentation team: https://github.com/github/docs/issues/14145.

PR #58, opened 2022-03-01
test: add debug prints for the SMTP server mock

Totktonada:
Usage:

```shell
$ DEBUG= ./test/smtp.test.lua
```

If the environment variable is set, it prints a log message, when a
client is connected, when the server reads or writes data and when the
server goes to close the socket.

Useful to understand what a client actually sees.


vr009 (approved):
Adding debug prints seems useful. Looks good to me.

PR #59, opened 2022-06-24
ci: use setup-tarantool action on Linux, add 2.10.0

Totktonada:
Using of the [setup-tarantool](https://github.com/marketplace/actions/setup-tarantool) action speed ups the testing using GitHub's caching infrastructure.

Also added testing on tarantool-2.10.0.


Totktonada:
Filed #60 about CentOS 8 fails. They're irrelevent here.

vr009 (approved):
LGTM.

PR #72, opened 2023-03-07
api: add hardcoded versioning support

GRISHNOV:
Add versioning support, now the module api has `_VERSION` hardcoded variable. Is part of the task [1].

1. https://github.com/tarantool/roadmap-internal/issues/204

### DIFF
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -3,12 +3,29 @@ name: packaging
 on: [push, pull_request]
 
 jobs:
+  # Run not only on tags, otherwise dependent job will skip.
+  version-check:
+    # Skip pull request job when the source branch is in the same
+    # repository.
+    if: |
+      github.event_name == 'push' ||
+      github.event.pull_request.head.repo.full_name != github.repository
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Check module version
+        # We need this step to run only on push with tag.
+        if: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags/') }}
+        uses: tarantool/actions/check-module-version@master
+        with:
+          module-name: 'smtp'
+
   build_package:
     # Skip pull request jobs when the source branch is in the same
     # repository.
     if: |
       github.event_name == 'push' ||
       github.event.pull_request.head.repo.full_name != github.repository
+    needs: version-check
 
     strategy:
       fail-fast: false
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -6,6 +6,16 @@ on:
     tags: ['*']
 
 jobs:
+  version-check:
+    # We need this job to run only on push with tag.
+    if: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags/') }}
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Check module version
+        uses: tarantool/actions/check-module-version@master
+        with:
+          module-name: 'smtp'
+
   publish-scm-1:
     if: github.ref == 'refs/heads/master'
     runs-on: ubuntu-latest
@@ -18,6 +28,7 @@ jobs:
 
   publish-tag:
     if: startsWith(github.ref, 'refs/tags/')
+    needs: version-check
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
--- a/smtp/CMakeLists.txt
+++ b/smtp/CMakeLists.txt
@@ -104,5 +104,5 @@ target_link_libraries(lib ${CMAKE_DL_LIBS})
 set_target_properties(lib PROPERTIES PREFIX "" OUTPUT_NAME "lib")
 
 # Install module
-install(FILES init.lua DESTINATION ${TARANTOOL_INSTALL_LUADIR}/${PROJECT_NAME}/)
+install(FILES init.lua version.lua DESTINATION ${TARANTOOL_INSTALL_LUADIR}/${PROJECT_NAME}/)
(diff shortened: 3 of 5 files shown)


.github/workflows/packaging.yml
```diff
@@ -3,12 +3,30 @@ name: packaging
 on: [push, pull_request]
 
 jobs:
+  version-check:
```
DifferentialOrange: You set two jobs here

.github/workflows/packaging.yml
```diff
@@ -3,12 +3,30 @@ name: packaging
...
+    # repository.
+    if: |
+      github.event_name == 'push' ||
+      github.event.pull_request.head.repo.full_name != github.repository
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Check module version
+        # We need this step to run only on push with tag.
+        if: ${{ github.event_name == 'push' && startsWith(github.ref, 'refs/tags/') }}
+        uses: tarantool/actions/check-module-version@igrishnov/gh-204-add-hardcoded-versioning-support
```
DifferentialOrange: Let's move to master

smtp/init.lua
```diff
@@ -277,7 +277,11 @@ curl_mt = {
 --
 -- Export
 --
-local this_module = { new = smtp_new, _CURL_VERSION = driver._CURL_VERSION }
+local this_module = {
+    new = smtp_new,
+    _CURL_VERSION = driver._CURL_VERSION,
+    _VERSION = require('smtp.version')
```
DifferentialOrange:
```suggestion
    _VERSION = require('smtp.version'),
```

GRISHNOV: Done

.github/workflows/packaging.yml
GRISHNOV: Done

.github/workflows/packaging.yml
GRISHNOV: Fixed

smtp/CMakeLists.txt
```diff
@@ -104,5 +104,5 @@ target_link_libraries(lib ${CMAKE_DL_LIBS})
 set_target_properties(lib PROPERTIES PREFIX "" OUTPUT_NAME "lib")
 
 # Install module
-install(FILES init.lua DESTINATION ${TARANTOOL_INSTALL_LUADIR}/${PROJECT_NAME}/)
+install(FILES init.lua version.lua DESTINATION ${TARANTOOL_INSTALL_LUADIR}/${PROJECT_NAME}/)
```
DifferentialOrange, line 107: Please, check that deb and rpm are fine

GRISHNOV: Checked for 
```
OS=debian DIST=bullseye packpack/packpack
OS=fedora DIST=35 packpack/packpack
```

The resulting `.tar.xz` contains a version.lua file

DifferentialOrange (approved):
MacOS CI is red, but it doesn't seem to be related to this PR.

Totktonada:
> MacOS CI is red, but it doesn't seem to be related to this PR.

I filed #74 regarding it.

smtp/version.lua
```diff
@@ -0,0 +1,4 @@
+-- Сontains the module version.
+-- Requires manual update in case of release commit.
+
+return '0.0.6'
```
Totktonada, line 4: Shouldn't it be 0.0.6-dev or kinda? There were many commits since the 0.0.6 release and scm-1 differs from it.

DifferentialOrange: We decided not to mark intermediate versions since it sounds rather bothersome to support. But it is recommended to make a new module release after this PR.

Totktonada: The following command will install a version, which reports itself as 0.0.6, but it is far beyond 0.0.6:

```
tarantoolctl rocks install smtp
```

It looks not okay to report 0.0.6 for something that is not 0.0.6. If the version reporting may produce incorrect result (from a user perspective), the feature has no sense. (And if it has no sense, it doesn't matter how much effort it requires.)

I would use maven-release-plugin approach instead, see [here](https://github.com/tarantool/tarantool-java/commits/master).

image

IOW,

* Write `0.0.7` into version.lua, tag this commit, make a release.
* Write `0.0.8-dev`, or `0.0.8-snapshot` or `scm-1` into version.lua, push as another commit.
* Continue development.

Totktonada: I engaged @LeonidVas into the discussion. Since we plan to make a release soon we agreed to proceed with current solution and return here back a bit later (discuss again). My concern is related to the scm-1 version is it seems acceptable to fix it later (because, unlike released versions, it is possible to fix later). I hope that not some many users will install the scm-1 version during this window.

.github/workflows/publish.yml
```diff
@@ -18,6 +28,7 @@ jobs:
 
   publish-tag:
     if: startsWith(github.ref, 'refs/tags/')
+    needs: version-check
```
Totktonada, line 31: Here a presence of the `if` clause means that the job will run disregarding whether the `version-check` succeeds if I understood the [documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds) correctly.

> If a job fails or is skipped, all jobs that need it are skipped **unless the jobs use a conditional expression that causes the job to continue.**

DifferentialOrange: It is not: the documentation describes the situation when `needs` overrided with `if always`, see this example run below

https://github.com/DifferentialOrange/queue/actions/runs/4471886246

Totktonada: It seems, it is specifically about `if: ${{ always() }}`. @DifferentialOrange shows me an example and it actually works (failed version check discards the following jobs).

(Update: I missed Georgy's comment, when I wrote this one. It appears that he already answered here.)

Totktonada (approved):

PR #73, opened 2023-03-18
Fix signed integer overflow in statistics

Totktonada:
A cast of the value above 2^63-1 to `int64_t` has implementation defined behavior, see n1256, 6.3.1.3.3. `lua_pushinteger()` accepts `int64_t`, so passing `uint64_t` value may lead to pushing negative (or other incorrect) value.

Let's use Tarantool specific `luaL_pushuint64()` function instead.

This is mostly just to make things in the right way: it is unlikely to see such large values in the statistics in practice.

See all the details in https://github.com/tarantool/tarantool/pull/8464

### DIFF
--- a/smtp/lib.c
+++ b/smtp/lib.c
@@ -55,7 +55,7 @@ static inline void
 lua_add_key_u64(lua_State *L, const char *key, uint64_t value)
 {
 	lua_pushstring(L, key);
-	lua_pushinteger(L, value);
+	luaL_pushuint64(L, value);
 	lua_settable(L, -3);
 }
 /* }}}


Buristan (approved):
Hi, Alexander!
Thanks for the patch!
LGTM!

Side note: I see that testing failed on macos-11/macos-12 [at this test](https://github.com/tarantool/smtp/blob/d29a025d3ab27f90c84fd4ee96ddbcd884b9fd38/test/smtp.test.lua#L273), with the error:
```yaml
expected: 'response reading failed (errno: 115)'
got: 'response reading failed (errno: 36)'
```
but it looks unrelated to the patch.

Totktonada:
Filed #74 regarding the fail. I agree, it looks unrelated.

PR #77, opened 2023-03-21
ci: workaround release rockspec publishing problem

Totktonada:
The problem is described in https://github.com/tarantool/setup-tarantool/issues/36: `tarantoolctl rocks` from tarantool 1.10 doesn't work on Ubuntu Jammy, when it is installed by the setup-tarantool GitHub Action. Let's use Ubuntu Focal as a temporary workaround.


Totktonada:
The problem appears here: https://github.com/tarantool/smtp/actions/runs/4474317681/jobs/7862667911.

GRISHNOV (approved):

PR #79, opened 2025-03-31
ci: bump actions to use ubuntu-24.04

patapenka-alexey:
Bump actions to use ubuntu-24.04 for fixing the following GitHub
warning:

    The Ubuntu 20.04 Actions runner image will begin deprecation
    on 2025-02-01.

Tarantool pre-build version has been increased to 2.11 because
`installer.sh` version 1.10 does not support Ubuntu `noble`.

Update CMake minumum version to prevent warnings about too old
CMake version during build configuring under Ubuntu 22/24.

Remove too old distros and unsupported tarantool versions from tests.

Part of #TNTP-1918

### DIFF
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -10,7 +10,7 @@ jobs:
     if: |
       github.event_name == 'push' ||
       github.event.pull_request.head.repo.full_name != github.repository
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     steps:
       - name: Check module version
         # We need this step to run only on push with tag.
@@ -32,18 +32,19 @@ jobs:
       matrix:
         platform:
           - {os: debian, dist: bullseye}
+          - {os: debian, dist: bookworm}
           - {os: ubuntu, dist: focal}
           - {os: ubuntu, dist: jammy}
-          - {os: centos, dist: 7}
+          - {os: ubuntu, dist: noble}
           - {os: centos, dist: 8}
           - {os: fedora, dist: 34}
           - {os: fedora, dist: 35}
           - {os: fedora, dist: 36}
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - name: Clone the module
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           # fetch-depth is 1 by default and it is okay for
           # building from a tag. However it is convenient to
@@ -51,7 +52,7 @@ jobs:
           fetch-depth: 0
 
       - name: Clone packpack
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           repository: packpack/packpack
           path: packpack
@@ -62,7 +63,7 @@ jobs:
           export DIST=${{ matrix.platform.dist }}
           ./packpack/packpack
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
           name: ${{ format('{0}-{1}', matrix.platform.os, matrix.platform.dist) }}
           path: build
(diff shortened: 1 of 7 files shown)


oleg-jukovec (changes requested):
Please, update macos runners (to macos-13). They are outdated too.

.github/workflows/testing.yml
```diff
@@ -17,50 +17,16 @@ jobs:
...
-          #
-          # Verify that system's libcurl headers are not required.
-          - tarantool: live/2.8
-            dont_install_system_libcurl_header: true
-          # Applicable for: 2.6.3, 2.7.2+, 2.8.1+, 2.9+ (gh-4559).
-          #
-          # Verify that tarantool's libcurl headers are preferred
-          # when the system provides them too.
-          - tarantool: live/2.9
-            break_system_libcurl_header: true
```
Totktonada: Why don't do the same against any modern tarantool version?

Removing these jobs effectively reverts 24c092053fbd6c4cf15514d30cedc6325c0850de.

.github/workflows/packaging.yml
```diff
@@ -32,26 +32,27 @@ jobs:
       matrix:
         platform:
           - {os: debian, dist: bullseye}
+          - {os: debian, dist: bookworm}
           - {os: ubuntu, dist: focal}
           - {os: ubuntu, dist: jammy}
-          - {os: centos, dist: 7}
```
Totktonada, line 37: There is no word in the commit message regarding this change. Something doesn't work? If so, what exactly? How it looks/how can I find that I met the same problem?

I don't mind the change, but please describe the motivation behind.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
```
Totktonada, line 1: Similarly, this is not described in the commit message. I suppose there is some problem behind. But only you know about it. Please, share it with others.

debian/prebuild.sh
```diff
@@ -1,4 +1,4 @@
-curl -LsSf https://www.tarantool.io/release/1.10/installer.sh | sudo bash
+curl -LsSf https://www.tarantool.io/release/2.11/installer.sh | sudo bash
```
Totktonada, line 1: Same, this change is not mentioned anywhere in the commit description. Does it fix some prolem? Or just hold the new minimal supported version?

CMakeLists.txt
patapenka-alexey: To prevent [such type](https://github.com/tarantool/checks/issues/64) of problems. Additionally described in the commit message.

debian/prebuild.sh
patapenka-alexey: `1.10/installer.sh` does not handle Ubuntu Noble, so I decided to update to newer version.

.github/workflows/packaging.yml
patapenka-alexey: Centos 7 is not officially supported since 2020.

.github/workflows/testing.yml
patapenka-alexey: I reverted this deletion and applied newest release versions.

.github/workflows/testing.yml
```diff
@@ -17,41 +17,23 @@ jobs:
...
-          - live/1.10
-          - live/2.6
-          - live/2.7
-          - live/2.8
-          - live/2.9
-        # Add an extra check or modify a system environment
+          - release/1.10.14
+          - release/2.11.4
+          - release/2.11.5
+          - release/2.11.6
```
oleg-jukovec: It looks like we need to add Tarantool 3.3.1 here instead of 2.11.4 and 2.11.5.

.github/workflows/testing.yml
```diff
@@ -121,27 +103,11 @@ jobs:
...
-          - 2.5.3
-          - 2.6.1
-          - 2.6.2
-          - 2.6.3
-          - 2.7.1
-          - 2.7.2
-          - 2.8.1
-          - 2.10.0
+          - 1.10.15
+          - 2.11.6
```
oleg-jukovec, line 111: We should add 3.3.1 here too.

oleg-jukovec (commented):
It would be nice if you split the changes into commits. Your commit now has 3 logical parts.

patapenka-alexey:
> It would be nice if you split the changes into commits. Your commit now has 3 logical parts.

I disagree. All these changes triggered by one reason - bumping Ubuntu version, splitting them into separate commits will lead to the red CI.

oleg-jukovec (approved):

